PR #94, opened 2019-01-06
fixed issues with IGraphics Mac dependencies build script

hor-net:
removed module name from version variable
changed mirror for freetype from savannah to freetype since the old one was not working anymore


AlexHarker:
The savannah issue is not reproducible here. Not sure whether having the module names at the top is a plus or a minus.

PR #129, opened 2019-02-13
Nimble Tools fixes

codecat:
Sorry for the large bunch of changes here - I'll make sure to organize it better in the future. This pull request features a number of changes across the board:

* Fixes crash on standalone `IPlugControls` example project (not sure why this happens actually)
* Updated `gitignore_template` to add additional Visual Studio and Xcode junk files
* Fixes text color in `IVKnobControl` not being set based on the color spec
* Fixes `IVMeterControl` to be a little more accurate
* Adds the ability to draw a frame around `IVScopeControl`
* Fixes crash when parameter is retrieved before it's attached to a delegate
* Adds a setter for the outer padding in `IVTrackControlBase`
* Adds a setter for drawing the frame of tracks in `IVTrackControlBase`
* Fixes "Live edit" not letting you move controls

If you really want, I can put these as separate pull requests, but it will take a little bit of time to put that together. (Again, I'll do better in the future.. :P)

### DIFF
--- a/Examples/IPlugControls/IPlugControls.cpp
+++ b/Examples/IPlugControls/IPlugControls.cpp
@@ -171,6 +171,10 @@ void IPlugControls::ProcessBlock(sample** inputs, sample** outputs, int nFrames)
 {
   const double gain = GetParam(kGain)->Value() / 100.;
   const int nChans = NOutChansConnected();
+
+  if (inputs == nullptr) {
+    return;
+  }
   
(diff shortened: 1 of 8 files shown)


olilarkin:
I'm going to close this because it's not possible to deal with non-atomic PRs like this. Please submit one thing at a time in discreet commits/feature PRs . I'll keep it in mind as i update vcontrols

PR #131, opened 2019-02-14
Fix HostPath on Windows, add INIPath function

ddf:
The INIPath implementation wound up being based on VST3PresetsPath but winds up generating the same folder path as the code in IPlugApp_host.cpp. I decided *not* to include the settings.ini filename in the path because IPlugAppHost::InitState wants the directory before it wants the filename so it can create the directory and file if they aren't there already. This way, if you decide to change InitState to use INIPath, it will look much the same as it does now.

### DIFF
--- a/IPlug/IPlugPaths.cpp
+++ b/IPlug/IPlugPaths.cpp
@@ -132,6 +132,13 @@ void SandboxSafeAppSupportPath(WDL_String& path)
   AppSupportPath(path);
 }
 
+void INIPath(WDL_String& path, const char * pluginName)
+{
+  GetKnownFolder(path, CSIDL_LOCAL_APPDATA);
+
+  path.AppendFormatted(MAX_WIN32_PATH_LEN, "\\%s", pluginName);
+}
+
 #elif defined OS_WEB
 
 void AppSupportPath(WDL_String& path, bool isSystem)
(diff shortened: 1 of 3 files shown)


AlexHarker:
A quick note to say that Oli's added CI tools that run checks and this PR is failing (see the GitHub page for the PR) - it looks pretty small, but ideally if you could trace this through and check that all the examples compile that would be really helpful as we are trying not to push anything to master that is broken.

ddf:
I am able to compile all of the examples except for IPlugFaust (due to some missing faust headers that IPlugFaust.h includes) and the ReaperExtension which spits out this error at the end of buildall-win.bat:

```
ReaperExtension.sln.metaproj : error MSB4126: The specified solution configuration "release|win32" is invalid. Please specify a valid solution configuration using the Configuration and Platform properties (e.g. MSBuild.exe Solution.sln /p:Configuration=Debug /p:Platform="Any CPU") or leave those properties blank to use the default solution configuration. [S:\dev\iPlug2\Examples\ReaperExtension\ReaperExtension.sln]
```

All of the errors appear to be related to SDK downloads failing and then compilation failing because the required SDK is not present. None of the errors are related to the code in the pull request.

olilarkin:
remade this PR to get CI to re-run, since CI results were useless at the time it was done.

closing this one

https://github.com/iPlug2/iPlug2/pull/147

PR #132, opened 2019-02-15
BREAKING - pass IParam::Shape by const reference and not pointer

AlexHarker:
This set of commits removes the need to use new when creating shapes for parameters, as shapes are passed by const-reference and the Clone() function is used to copy the shape object, leading to nicer code.

@olilarkin  - I'm hoping you'll like this change, so I'm making a PR because of the breaking nature of the change only - I've tested - it works and most of the changes are simply to follow through the lower level change.


johannesmenzel:
@AlexHarker I really appreciate that kind of changes. When i find time, i'd pull my custom SetShapeFromMiddlePosition Shape, if you could use it.

AlexHarker:
No problem - please propose it if you think that it could be of general use.

PR #136, opened 2019-02-21
Fix SVG hole/solid detection with NanoVG 

madronalabs:
I added a tiny Vec2 object to IGraphicsStructs.h/IGraphicsUtilities.h to implement the algorithm. I couldn't find a Vec2 class anywhere in IGraphics, which seems weird, so maybe I missed something.


olilarkin:
@madronalabs have you got an example svg to test this with?

madronalabs:
see https://github.com/iPlug2/iPlug2/pull/163 for example.

AlexHarker:
Hi Randy,

I think I understand all this now, but I'm worried about how it fits with drawing manually or with other drawing APIs that behave differently. I'd prefer it if we could address NanoVG similarly to other APIs,

Short of maintaining and transforming our own stack of commands above NanoVG and then doing the work you are doing in the SVG render for that stack (which would be a pain), I can't see how we do this without modifying NanoVG.

If we took this second route your code would mostly go into:

nvg__flattenPaths

I'd suggest having an option where NanoVG respects user windings, and another whether it does winding/evenodd behaviour according to a user setting. I'm not sure what your thoughts are on either of these two approaches? We have our own fork of nanovg that has some font loading stuff in it (I hope that will get pulled into the main repo from a recent PR). This is a bit lower level than that, but it is one option.

@olilarkin - do you have any thoughts about modifying NanoVG/building on top of it? If we don't do either of these then there is a clash between the flow in different APIs. I know old likes NanoVG, but it does have limitations that we don't see in Cairo etc. (Mac OS / issues with gradient rendering on some systems / lack of multipoint gradients / no line dashes).

olilarkin:
I have no problem to use our own NanoVG fork, if any modifications are in sensible pull requests  to the main project. NanoVG is looking for maintainers https://github.com/memononen/nanovg/issues/502. I feel like I don't have the graphics chops to do  any heavy lifting there ( maybe one of you does?).

AlexHarker:
I *think* I could apply Randy's fix to NanoVG, although he might be able to do it better, or to check it when I'm done.

AlexHarker:
OK - I've looked at this and I've run into a few issues:

1 - the main loop appears to read beyond the set of points

2 - within nanoVG once the paths are "flattened" (the easiest time to operate on them) the bezier curves are also "tessellated" which as I read it turns them into line segments (leading to a lot more tests potentially needing to be run)

3 - the algorithm is for even odd only (not winding behaviour as used in Cairo/AGG/Skia/Canvas)

@madronalabs - some input here would be really useful.

madronalabs:
1 - Though the loop is written in a confusing way I don't think it does.

2 - I don't understand what you mean here. What difference in the code are you proposing?

3 - With differing winding rules, certain SVGs with intersecting paths will look different in the different renderers. I don't see this as a problem in practice because I would never generate such an SVG in production. 

I don't see how this fix falls into the realm of nanovg at all, given that nanovg does not have an svg renderer. Its own path-filling behavior seems completely fine though it differs from some of the other renderers. In other words, it's not broken. I feel like it's a good compromise between utility and speed. 

The winding correction in this commit should produce the same results on all the renderers for all SVGs that don't have intersecting paths.

Where identical visual results for different renderers is a requirement, adding some additional helper functions "NanoVGHelpers.h" seems like a fine approach. You would only have to wrap any renderers that do require such tweaks.

To focus discussion about these rendering issues I think having a canonical *static* test pattern would be a good idea. It would be define the project's graphical requirements: that this block of code with these input media (SVG, raster images) must produce this raster output, give or take a few pixels. There could even be automated tests of such.

AlexHarker:
The pattern we have at the moment for IGraphicsTest is that for fills you draw paths (with subpaths) before calling fill. When you call fill you pick the winding mode for what actually get filled. Those two modes (winding and even odd) are modelled on on the graphics libraries that use this common pattern.

NanoVG doesn't work that way - it requires you to manually set each subpath whilst drawing - which can't be matched easily with any other drawing API.

For me this is not an SVG rendering specific issue, because if I was to generate all the relevant calls by hand (as a series of PathMoveTo / PathCurveTo / Fill calls) in Cairo/Canvas it would draw the shape correctly. SVGs with a fill rule for winding (not even odd) also would be rendered as even odd in NanoVG with this fix.

I was proposing a mod to NanoVG which I attempted - This is where I got to:

https://github.com/iPlug2/nanovg/tree/even-odd-winding 

But, as this code is not always receiving curves, and I am not really sure of the details of how it *should* work, I couldn't get it 100% working.

The relevant functions are:

nvg__pathsWinding.     and
nvg__getLineCrossing

There are now two items in IGraphicsTest that render with even odd winding (the SVG you provided and the multipath at the top right). If the fix is successful both of those will render with holes.

AlexHarker:
Just noticed that if I change the loop to work on line segments then it works, but it is a lot of tests given that the paths now have maybe 32/64 points per curve.

AlexHarker:
This iPlug branch shows something that seems to work most of the time in all cases:

 https://github.com/iPlug2/iPlug2/tree/nanovg-winding-tests

I have no idea if I'm doing the core testing correctly, although the tests I've run come out correctly.

AlexHarker:
I added some tests to IGraphicsTest on the multi-path control but an added complication is that for a crossed path Cairo behaves differently to NanoVG. The paths are taken from here:

https://oreillymedia.github.io/Using_SVG/extras/ch06-fill-rule.html

When the controls are less than half the fill rule is nonzero and above that even-odd.

madronalabs:
I see what you mean about this not being an SVG-specific issue. I had forgotten that. 
I also see how my PR adds complexity for other renderers by adding checks in DrawSVG(), which is obviously bad. 

The change to nanovg looks reasonable. Oli is familiar with that project so he may have some idea whether such a change is likely to be accepted. 

Meanwhile could we put the winding code into IGraphicsNanoVG::PathFill() instead?  

I got your  nanovg-winding-tests branch to compile — the svg looks good, but the TestArcControl is broken now. It appears to be filling the whole bounding rectangle of its arc with blue:

Screen Shot 2019-04-08 at 9 31 52 AM

AlexHarker:
PathFill would be a good place to put this, but unfortunately you can only set the current subpath winding in nanovg and by the time you fill the path all but the last subpath has already been set in nanovg.

I'll check out the winding branch again - I didn't notice an issue with the arc control here, but I'm also unsure exactly how to proceed as a true even/odd algorithm would allow holes for a path that crosses itself, but this doesn't appear possible in nanovg. 

One issue is that I don't really understand the path winding tests you wrote as compared to the issue of whether a given pixel should be filled (which is the test a rasteriser does) and most internet sources deal with the latter, so I'm not sure how the non-zero and even odd winding rules apply to those tests. Any pointers?

olilarkin:
was looking at the VCVRack svg code and noticed the similarity to what randy did. Updated version of his branch here https://github.com/iPlug2/iPlug2/tree/graphics/svg-compound-shapes

Doesn't have an adverse affect on SVGs in other backends as far as I can tell.

olilarkin:
moved to #377 with new PR

Issue #142, opened 2019-03-08
macOS 32 bit build fails in IGraphicsMac_view.mm


hor-net:
When building the 32bit version of mac plugin the build fails with a c++11-narrowing error thrown on lines 723 and 724 of IGraphicsMac_view.mm

the two offsets are defined as double, the should be CGFloat instead, this fixes the problem.

olilarkin:
can you make a PR? i can't compile 32 bit any more

AlexHarker:
sounds like a warning that is being promoted to an error.

hor-net:
Sent the pull request:  
https://github.com/iPlug2/iPlug2/pull/143

AlexHarker:
I fixed this slightly differently using static_casts - can you test?

olilarkin:
closing this as i am fairly sure Alex's fix will have solved it

AlexHarker:
Thanks Oli - I did test against 32bit here and it made it work.

PR #143, opened 2019-03-11
Igraphics 32bit cgfloat

hor-net:
Pull request that fixes issue #142


AlexHarker:
Fixed with slightly different code

PR #145, opened 2019-03-18
Win faust support

dmsplugins:
This is just a step in the right direction.... I think. The path definitions are just for testing. A better solution is needed.


olilarkin:
I'm going to close this. I have started work on it here...

I am speaking to the faust guys and i hope they can provide a dll with the faust installer, which should make things simpler

https://github.com/iPlug2/iPlug2/tree/faust-win

PR #151, opened 2019-03-21
Add support for JIT compiling with libfaust on windows

olilarkin:
Currently requires a specially built .dll in ProgramFiles/faust/lib


olilarkin:
closing as some wrong commits included

PR #154, opened 2019-03-24
Add demo control using Metal Performance Shaders on macOS/iOS

olilarkin:


olilarkin:
hmmm DrawFittedLayer is not correct wrt draw/screen scaling

PR #156, opened 2019-03-26
Cairo Font loading for Windows

ddf:
Not included in these changes are some small modifications to IPlugControls for testing loading a font from disk using a Bold style. I could push these and update the pull request if you'd like, but I wasn't sure if that was the most appropriate example for them.

Notes on these changes:

* using `get_filepart()` to extract only the name of the font (eg Verdana, Roboto-Regular, etc) to use as the key for the font cache because otherwise fonts loaded using a full system path wind up with the path as part of the lookup key, which causes FindFont to fail, since the pattern with IText is to put the name of the font without a path or extension in the mFont field
* on Windows, if OSFindResource can't locate the font, we still attempt to create an HFont with the name provided because Windows is able to load fonts installed on the system by name (eg Verdana.ttf)
* changed FindFont to search for the font-style name first because otherwise attempting to find Bold and Italic fonts will always return the Normal face if that has also been loaded

Notes from testing:

* iPlug is not automatically loading DEFAULT_FONT, which meant that even after adding font loading support for Cairo, I also had to add a `LoadFont("Verdana.ttf")` call in order for IPlugControls to run properly. This should probably not be necessary? For example IPlugControls will assert in FindFont without adding code to explicitly load Verdana.
* if a font fails to load, then text using that font will simply not appear, which is different from the behavior on master where it winds up falling back to some default system font
* macOS is not able to load system fonts with this current code, but I did find that `CGCreateFontWithName` will work if provided a system font name without the extension (eg Helvetica Neue). Related: DEFAULT_FONT for macOS is current "Verdana", which is not a stock installed system font. This should probably be "Helvetica Neue"

Aside from that, was curious why the need for two LoadFont methods, rather than one with the `IText::EStyle` argument defaulted to `IText::kStyleNormal`.


ddf:
Ok, I see now what the difference between the two font loading methods is supposed to be. Is it *intended* that `LoadFont(const char*, IText::EStyle)` should fail if full path to a font file is provided?

ddf:
You know what, I don't think these changes are correct. From doing some more research, it seems that CreateFont can return a font that has similar characteristics to the one provided. I will work on this more an update the pull request.

AlexHarker:
The intention is the one loads with a file path. The other loads *any available system font* without an actual path. I know about the default font problem - this needs to be discussed with Oli.

the Mac OS code *does load system fonts* but you shouldn't be using paths. So:

LoadFont("Baskerville", IText::kStyleNormal);

should give you Baskerville. The type of the system font shouldn't matter (ttf/ttc/otf etc.) because CoreText is doing the work for us.

AlexHarker:
Thanks for looking at this BTW.

ddf:
Cool, yep, that's where I landed after further review and I'm working on changes that address this.

ddf:
Roboto-Regular.ttf needed to be renamed to Roboto.ttf in order to get this workings because the file path or loaded resource is used to to "install" the font for the process and then `CreateFont` can be called with the font family name. So, if the file name is not the same as the font name as listed inside of the ttf file, then `CreateFont` will not choose the correct font when used in `LoadFont(const char* name)`.

See: https://www.codeproject.com/Articles/42041/How-to-Use-a-Font-Without-Installing-it

We could bring in code for parsing the ttf header in order to find the name of the font instead of calling `CreateFont` with the trimmed file name (nanovg's font handling code parses this info fwiw), but that would break the convention that `IText::mFont` is always the same as the font file name without the ttf extension. The thing is, if we have users follow the convention that ttf files in Resources are named the same as the font name, then that means fonts loaded from Resources will *also* be used in Windows platform text entries, since those call `CreateFont` with `IText::mFont` in order to set the font on the text edit, which is not something that has been supported.

ddf:
A thing I still need to test: the article above and the docs for `AddFontResourceEx` say that any fonts added for a process are uninstalled when the process ends, so I'm not sure if the font caching as written will work properly on Windows. Eg: two instances of a plug using a Resources font are loaded in a DAW, then the user removes on of the plug instances, will the remaining instance start rendering the bundled font incorrectly?

PR #161, opened 2019-04-03
Simplify CloseWindow() logic and try to avoid re-entrancy issues

### LINKED ISSUE #160: Crashes on application quit with nanoVG
When I quit any plugin or app made with iGraphics / nanovg I'm getting frequent crashes.

IGraphicsMac::CloseWindow() calls [view removeFromSuperview] which in turn calls IGraphicsMac::CloseWindow() again. Can this be right? It makes this code possibly harder than necessary to reason about. It also looks like there is potentially a race condition on mView.

The crash occurs in IGraphics::RemoveAllControls(), where it looks like the IGraphics object is in the process of being freed because the member variables point to garbage.

AlexHarker:
This attempts to fix #160. I still don't understand exactly the conditions that cause the crash, but it seems to be re-entrancy related. I can't see any race conditions (there is only one thread involved), but I there is an outside chance it is memory corruption/stamping.

What I've done here is:

1 - stop IGEditorDelegate::CloseWindow from running code twice if it does re-enter to stop the crash
2 - fixed a memory leak where mGraphics wasn't deleted for non-transient graphics.

I was getting the crash intermittently with iPlugEffect. Now if I load 10-20 times in a row and quit I get no crash.

### DIFF
--- a/IGraphics/IGraphicsEditorDelegate.cpp
+++ b/IGraphics/IGraphicsEditorDelegate.cpp
@@ -19,6 +19,8 @@ IGEditorDelegate::IGEditorDelegate(int nParams)
 
 IGEditorDelegate::~IGEditorDelegate()
 {
+  if (mGraphics)
+	DELETE_NULL(mGraphics);
 }
 
 void IGEditorDelegate::OnUIOpen()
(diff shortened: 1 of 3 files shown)


madronalabs:
Great, I'll update and keep you posted.

PR #162, opened 2019-04-04
Move OSFindResource() and related methods to IPlugPaths

olilarkin:
This also removes ToNSString and CocoaAutoReleasePool


olilarkin:
some overlap here with objective-c-mem-management branch

olilarkin:
GetResourcePathFromSharedLocation needs to change -  passing in a globally set location for shared resources, currently has a hardcoded path

PR #163, opened 2019-04-05
added test SVG with compound shapes to IGraphicsTest

madronalabs:
If you merge this PR, a new test SVG with holes should appear in the IGraphicsTest instead of one of the tigers. This will draw incorrectly. Then adding the "added hole/solid detection" fix should cause the letters and shapes to draw properly.


AlexHarker:
Thanks - this is super clear. It works fine in AGG, so obviously this is a NanoVG only issue. I'll take a look at the fix, with a view to merging either as is or with some mods for design consistency.

PR #166, opened 2019-04-06
Fix CompileCPP for Windows Faust Support

dmsplugins:
Seems to work ok, but I found it necessary to delete FaustCode.hpp and let CompileCPP make a new one.
By the way, the stuff you've done to improve IPlugFaust is really cool.


olilarkin:
Thanks!

dmsplugins:
In my duplicated project, I additionally moved the DSP_FILE definition from EXTRA_ALL_DEFS to EXTRA_DEBUG_DEFS and changed it to correspond to my folder structure. Would it be appropriate to do the same to the project in the Examples folder with instructions for users to manually modify the property, or is there  a better way to define DSP_FILE?

PR #169, opened 2019-04-07
Bitmap button/switch control constructors clip bounds to bitmap

drakfluga:
Bitmap button/switch now clips its bounds to the bitmap dimensions similarly to knobs.


drakfluga:
Before:

incorrect

After:
correct

olilarkin:
Seems sensible. These controls and the demo was never finished... as you saw in the screenshot

PR #171, opened 2019-04-07
duplicate projects to different location

pressplay123:
Duplicating an iplug project works only within the same path. If you move a duplicated project to another location, you have to correct project files (header, framework, script locations...) manually, which can be quite frustrating. I have changed the duplicate.py script from the Examples folder, which now takes an optional argument for the output path. Like this:

./duplicate.py IPlugEffect NewEffect pressplay ../../../MyPlugins

It works with absolute paths, too. But it will probably only work if the output path is outside of the iplug2 folder structure (but that's the use case, isn't it?)

I had to change more than I originally planned :( And I did not test on Windows.

I did not change the makedist-... scripts (too many of them), but this should be straightforward:
IPLUG_ROOT=../../..
cp ${IPLUG_ROOT}/.. ......... etc

Can you please check if it works correctly and if it perhaps breaks things I was not aware of?


pressplay123:
Not sure what those errors mean. Incompatibility with your integration tests?

olilarkin:
thanks for the PR...

> Not sure what those errors mean. Incompatibility with your integration tests?

Continuous integration failed, I think it's because you didn't modify the python scripts  for the projects in the "Tests"  folder

pressplay123:
Yes, I did not touch the test folder. Will do this later or maybe tomorrow.

drakfluga:
I'll follow this! I try to separate my own code from libraries etc, it would be great to be able to plop my project down outside the IPlug folder hierarchy. :)

pressplay123:
On Windows, there is this backslash thing 🙄 so please wait for another commit...

pressplay123:
okay, it works on my side and I believe there is nothing left to do. Hope you find it useful. This was my first pull request on github, yeah

olilarkin:
thanks for this, will merge soon when i have had time to test it a bit

olilarkin:
closing this one... i made a few more commits to change ROOT -> IPLUG2_ROOT and to, in theory make it work for Web make files too (although there is more work to be done in the build scripts for that). I couldnt work out to push to this PR to check CI, so i made a new PR #189

olilarkin:
works well on mac (apart from lots of scripts needing updating like you said), Does not seem to work on windows, in that all the relative paths in the vsprojects didn't get changed. Work will continue on this branch https://github.com/iPlug2/iPlug2/tree/duplicateoutofsource

PR #172, opened 2019-04-08
Added HiDPI test graphics to IPlugControls project

drakfluga:
I changed the knob PNGs in the test project to display "1x" or "2x" depending on scale. Could be rolled into master or only kept in this branch for testing if you like.


olilarkin:
seems useful, will think about it when merging graphics-winhidpi to master

Issue #47, opened 2018-11-11
IPlugVST3/IPlugVST3Controller/IPlugVST3Processor  should share more code


olilarkin:
There is a lot of duplication between the distributed and non-distributed VST3 code

AlexHarker:
This is substantially addressed on the vst3-changes branch. There is more to do, and some file structuring/class naming, but there is far less duplication than before. I think the processor common code is probably done, but the controller needs looking at.

PR #174, opened 2019-04-09
VST3 base classes sharing more file code

olilarkin:
Merging this slightly prematurely to reduce the pain of updating projects


olilarkin:
closes #47

Issue #122, opened 2019-02-08
WAM examples throw exception when browser zoom != 100%


olilarkin:
refresh page with zoom != 100% = exception

AlexHarker:
Depends on the size - this appears to be an issue with bitmaps where the scale is beneath 50% - I suspect we need a minimum scale of 1 for finding images.

AlexHarker:
Will also depend on the screen resolution. On a retina this will happen under 50%. On non-retina anything under 100%.

I've fixed it on master, but the issue stems from devicePixelRatio (retrieved in OpenWindow()) incorporating the zoom amount. We should decide how this should work, because right now different bitmaps will be loaded according to the zoom on open, but zooming after loading won't change the bitmaps.

PR #181, opened 2019-04-12
Fix issues with screen scale on web by checking for changes on each t…

AlexHarker:
…imer call.

this should fix up the remaining issues with #122.

Zoom is now taken into account in screen scale, but that is probably no bad thing.

### DIFF
--- a/IGraphics/Controls/Test/TestFontControl.h
+++ b/IGraphics/Controls/Test/TestFontControl.h
@@ -68,7 +68,7 @@ public:
     mStrCount = 1 - mStrCount;
       
     IColor c = DEFAULT_TEXT_FGCOLOR;
-    const char* font = mFontCount ? "Roboto-Regular" : "Times New Roman";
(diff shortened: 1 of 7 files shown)


olilarkin:
This is working well on my non-retina screen, but on the retina screen with IPlugControls,  the bitmap frames are incorrect when I zoom in
Screen Shot 2019-04-13 at 13 45 06

AlexHarker:
Can you check that I'm correct that this doesn't happen with canvas?

The issue here is to do with what happens on control resizing. With nanovg the bitmaps don't scale, but my guess is that they are stored as if they do. 

If you build on master zoom and reload the page with nanovg you'll see the same issue, so this PR is merely uncovering another issue, rather than creating it directly.

AlexHarker:
OK - actually it was how NanoVG was dealing with drawing bitmaps that weren't at the right size. Now fixed.

olilarkin:
seems to fix it

Issue #88, opened 2019-01-02
WebGL context should be done via html5.h, rather than emscripten GLFW emulation


olilarkin:
IGraphicsNanoVG::OnViewInitialized is a bit of a mess. Possibly OpenGL/Metal context wrangling should move to platform class.

https://github.com/iPlug2/iPlug2/blob/master/IGraphics/Drawing/IGraphicsNanoVG.cpp#L402

Anyhow...don't really need to use GLFW, which is wasteful, since not using a lot of that functionality.

better to do it with html5 WebGL stuff..

https://kripken.github.io/emscripten-site/docs/api_reference/html5.h.htm

PR #182, opened 2019-04-12
Enable macOS OpenGL rendering via NSOpenGLLayer

olilarkin:
This PR allows macOS to use a layer backed OpenGL view for nanovg (and skia) 

It also tidies up the NanoVG WebGL context creation code and adds an OpenGL test control


olilarkin:
hmm not scaling properly on retina

AlexHarker:
Looks like this should cover #21 and #88. Is the issue with scaling on desktop or web (or both)?

olilarkin:
Desktop, web is okay

AlexHarker:
OK - does it just not scale for the screen?

AlexHarker:
Isn't #21 now closable?

PR #189, opened 2019-04-17
Pressplay123's "extend duplicate script"

olilarkin:
This adds preliminary functionality to duplicate a project out of the IPlug source tree, useful if, like many do, you want to commit IPlug and Project files to SCM together


sstillwell:
In our internal IPlug, practice is to use WDL/IPlug as a submodule of the plugin project.  That way IPlug can be maintained separately, yet still be tracked to the individual commit within the separate plugin project.  Is there anything sub-optimal with that?

olilarkin:
Nope. This makes that possible/much easierf using the duplicate script

olilarkin:
closing this PR because duplicating out of source doesn't work yet on windows

pressplay123:
Strange, it works for me. I will have another look soon.

PR #190, opened 2019-04-17
Vector triangle control

drakfluga:
I made a drawing function for a triangular button (modelled on DrawVectorButton() ) and a control to go with it. I find it quite useful for creating up/down, increase/decrease value button pairs. The drawing function is the meat of this whole thing; the control may be less useful as we lack the concept of child controls - I included it more as a proof of concept.

Rotating the control only works for right angles as of now. It should be fairly straightforward to fix so that the control resizes to fit the bounds etc, but I'd say 99% of the time you only want right angles anyway.

If you accept this I'll be happy to add demo code to the controls example project.


drakfluga:
Oops, the implementation of SetColors() went along with the commit. Feel free to disregard it or include it if you want.

PR #193, opened 2019-04-17
Vector control styling & customization

drakfluga:
Added an IVStyle struct and corresponding IVectorBase constructor to facilitate easy customization of vector controls.

This can be safely merged and shouldn't break old projects but I'll be changing the example project(s) to demo styling and pushing it to this branch.

For now I've left IVColorSpec separate from IVStyle, color themes etc should be more easily supported if they're separate, but it does make the code slightly more verbose.

I closed my previous PR regarding the triangle button, that's included here as well.


drakfluga:
Pushed a few commits that fix embossed vector button drawing and uses IVStyle in the example project.

olilarkin:
This is good stuff, but not exactly how i want to do it. Closing this, but implementing some of the ideas on this branch https://github.com/iPlug2/iPlug2/tree/graphics-vcontrols

PR #209, opened 2019-04-29
Update windows projects for VS2019

olilarkin:


olilarkin:
closing PR since i think it is not VS2019 time just yet

PR #212, opened 2019-05-02
Fixed white background while resizing on Windows

Youlean:


Youlean:
> Makes it somehow better. But causes strange repaint behavior, e.g.:
> 
> * the scaled screen is not painted properly at the top left corner
This is probably caused because resizing is invoked by the mouse movement. Ideally, you should mark GUI for resize and do it on the next GUI event. 

> * the resizing is not as smooth as it is without the fix
It is not since it will force repainting the GUI, otherwise, it will be fast but you will see the white background.

olilarkin:
Although it's not perfect, this does improve it a lot - it makes it flicker less. The main issue that I see is that there is a significant lag (the IGraphics context does not become the new size fast enough). Merging, will try and improve this behaviour at some point

PR #213, opened 2019-05-04
fix setting the background color in a native windows text entry

ddf:
We need to SetDCBrushColor and then return the DC_BRUSH in order for the background color of parts of the text entry not behind actual text to be the correct color.

Docs for handling WM_CTLCOLOREDIT mention that if the message is handled, it must return the handle to a brush. See: https://docs.microsoft.com/en-us/windows/desktop/controls/wm-ctlcoloredit


AlexHarker:
Do we have any tests for this, or examples of realworld scenarios where you see the issue without the fix? Seems like you've looked into this in detail, but I'm confused that In IPlug1 setting background color on edits worked (as far as I remember) because I was using it....

ddf:
IIRC this was also broken in iPlug1, but whether you see it or not depends on how big the IRECT for the text entry is compared to the text size and whether or not the background color for the text edit is changed from the default value of white. 

Easiest way to see this is to run the IPlugControls example with CAIRO and use native text entry on Windows. Setup the ICaptionControl at line 159 with an IText that has a font size of 30 and background color of red. When you click on the number to edit it you'll see the background change to red only for the height of the text and the rest of the rect will be white.

ddf:
I made a similar fix in my fork of iPlug1, which I don't think was ever merged back:
https://github.com/ddf/wdl-ol/blob/7e5d1602e38b95f587870467b8a93e5962370b1c/WDL/IPlug/IGraphicsWin.cpp#L375

AlexHarker:
Thanks - I'm convinced, so I'll merge.

PR #214, opened 2019-05-06
IGraphics::PathArc winding parameter

ppryamikov:
Arc direction is needed when it is a part of combined path.

### DIFF
--- a/IGraphics/Drawing/IGraphicsAGG.cpp
+++ b/IGraphics/Drawing/IGraphicsAGG.cpp
@@ -279,11 +279,11 @@ void IGraphicsAGG::DrawBitmap(const IBitmap& bitmap, const IRECT& dest, int srcX
   }
 }
 
-void IGraphicsAGG::PathArc(float cx, float cy, float r, float aMin, float aMax)
+void IGraphicsAGG::PathArc(float cx, float cy, float r, float aMin, float aMax, EWinding winding)
 {
   agg::path_storage transformedPath;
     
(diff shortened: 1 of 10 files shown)


AlexHarker:
Thanks for pointing this out. Do you have a test case for this issue? There are currently some winding issues in general to do with NanoVG so I'd like to look a this all together if possible.

BTW - at the moment the CI is failing because IGraphicsCanvas hasn't been updated.

ppryamikov:
I've tested it on this code
```
auto centerX = mRECT.MW();
auto centerY = mRECT.MH();
auto radius = mRECT.W() * 0.25f;
auto width = radius * 0.75f;
auto startAngle = -90.0f;
auto endAngle = +90.0f;

g.PathArc(centerX, centerY, radius - width * 0.5f, startAngle, endAngle);
g.PathArc(centerX, centerY, radius + width * 0.5f, endAngle, startAngle, kWindingCCW);
g.PathClose();
g.PathFill(IPattern(COLOR_WHITE));
```
which should draw a shape like this if everything is fine
image
and all kinds of other things if winding of either arc is wrong.

Though I've tested it only with NanoVG yet.

ppryamikov:
Now tested it with Cairo and AGG. Fixed AGG which have misnamed parameter or different angles apparently.

Haven't tested with Canvas because not so much interested in it and also can't find info on how to build it. But if you're interested in pulling this I can take a look.

ppryamikov:
Now tested it with Cairo and AGG. Fixed AGG which have misnamed parameter or different angles apparently.

Haven't tested with Canvas because not so much interested in it and also can't find info on how to build it. But if you're interested in pulling this I can test it too.

PR #220, opened 2019-05-19
Replace typedef with using, closes #219

### LINKED ISSUE #219: replace typedefs with using ... =

olilarkin:


AlexHarker:
In this version you've changed one of the AGG types to Alpha_span_generator, with a leading capital. I put it in snake case to match AGG style - I don't mind whether it is snake or camel, but to me a single leading capital on snake is neither, so should we update to make consistent?

PR #221, opened 2019-05-20
Fix Win build for IPlugReaperExtension

dmsplugins:
Just some project settings.... BTW all example and test projects currently build on vs 2017 & 2019


olilarkin:
addressed in  #238, thanks

PR #222, opened 2019-05-22
VST2 key events

tommitytom:
Adds support for receiving keystrokes in VST2 plugins in certain hosts (like Ableton Live).  A few things to note:

* Only tested on Windows
* As discussed on Slack,  `OnKeyDown(const IKeyPress& key)` and `OnKeyUp(const IKeyPress& key)` were added to `IPlugAPIBase.h`.  I predeclared `IKeyPress` to avoid an `IGraphics*.h` include.
* I wasn't able to do the same trick for `IPlugVST2.cpp` base, however, as that required the VK enums, and `IKeyPress`.  I wasn't sure whether to further abstract those two things further away from graphics, but they could go in to something like `IPlugKeyboard.h`.
* VK's are mapped from the VST's own keycodes.  The VST keycodes are nowhere near as vast as the windows ones, so to fill the gaps I've implemented an `AsciiToVK(int ascii)` function.  On Windows there is a win32 api call to handle the conversion, taking in to account the keyboard layout/language.  On other platforms I've just mapped the alpha numeric keys to VK's since they pretty much match.  I'm not sure how to handle the mapping of other keys without keyboard layout information.  `AsciiToVK` doesn't belong where it is, espceially since it contains platform specific code.  It would be good if you could let me know where it belongs (maybe as part of `IPlugKeyboard` mentioned earlier?).
* My editor is set to remove whitespace at the end of lines when I hit save, and it seems someones editor puts in spaces on empty lines, which has resulted in lots of changes to empty lines!  I initially thought this was a line endings issue, and was racking my brains trying to fix it until I realise what was going on.  I hope this isn't an issue?  Do those spaces serve a purpose? :P
* I added key forwarding to the IPlugInstrument example.  I wasn't sure what to supply to the x,y args, what is best practice here?


olilarkin:
thanks for this. Sorry it took a while. I reworked at bit, see PR #330

PR #253, opened 2019-07-11
Auto switch to image surface

hor-net:
Choose an image surface instead of a quartz one at runtime if for the first second of rendering the frame rate is not as required in PLUG_FPS in config.h because on some Macs image surface is actually faster than CoreGraphics

### DIFF
--- a/IGraphics/Drawing/IGraphicsCairo.cpp
+++ b/IGraphics/Drawing/IGraphicsCairo.cpp
@@ -592,13 +592,12 @@ void IGraphicsCairo::SetPlatformContext(void* pContext)
   else if(!mSurface)
   {
 #ifdef OS_MAC
-#ifndef IMG_TEST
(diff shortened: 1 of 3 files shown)


AlexHarker:
I had intended just to allow the user (developer) to select image or quartz. The assumption here is that image is faster than quartz, so if a complex UI is rendered too slowly on a machine where quartz is faster, and that render doesn't return in time the backend will switch to image, which will then make the render even slower.

Doing a speed test with unknown drawing seems unwise.

hor-net:
oh, so we started from different assumption here, switching to IMG with a compile flag it's even easier I think, I'm a little busy right now, once done i'll send you a new PR

AlexHarker:
I don't mind the switching being run-time, but I don't think you should make it switch automatically...

hor-net:
oh ok so a use quartz/image surface method somewhere 👍

PR #262, opened 2019-07-18
new few docs, you find my comment below

enricocupellini:
1) not clear the trackNames in IVMultiSliderControl.
2) I edited last documentation merge because it resulted not well formatted on the web page 
3) I think the proposed way to have a feedback from resulting IVMultiSliderControls is not the best (requires too much passages)

According to the chosen constructor, I succeeded in get feedback from the slider in this way, but I think is not the best. How can I use the slider differently?

MyProgect:: MyProgect(IPlugInstanceInfo instanceInfo)
: IPLUG_CTOR(kNumParams, kNumPrograms, instanceInfo)
{
   ...
   pGraphics->AttachControl(new IVMultiSliderControl<4>(myBounds, "", DEFAULT_STYLE, EDirection::Vertical, -1., 1.));
      dynamic_cast<IVMultiSliderControl<4>*>(pGraphics->GetControl(pGraphics->NControls()-1))->SetParamIdx(kTag, 0);
      dynamic_cast<IVMultiSliderControl<4>*>(pGraphics->GetControl(pGraphics->NControls()-1))->SetTag(kTag);
...
}

void MyProgect::OnParamChange(int paramIdx) {
    switch (paramIdx) {
 ...
        case kTag:
        {
            if (GetUI() != nullptr) {
                IControl *c = (GetUI()->GetControlWithTag(kTag));
                printf("%f", c->GetValue());
            }
...
}

### DIFF
--- a/IGraphics/Controls/IControls.h
+++ b/IGraphics/Controls/IControls.h
@@ -168,24 +168,24 @@ class IVRadioButtonControl : public IVTabSwitchControl
 {
 public:
   /** Constructs a vector radio button control, linked to a parameter
-   * @bounds The control's bounds
-   * @paramIdx The parameter index to link this control to
-   * @label The label for the vector control, leave empty for no label
-   * @style The styling of this vector control \see IVStyle
-   * @shape The buttons shape \see IVShape
-   * @direction The direction of the buttons
-   * @buttonSize The size of the buttons */
+   * @param bounds The control's bounds
+   * @param paramIdx The parameter index to link this control to
+   * @param label The label for the vector control, leave empty for no label
+   * @param style The styling of this vector control \see IVStyle
+   * @param shape The buttons shape \see IVShape
+   * @param direction The direction of the buttons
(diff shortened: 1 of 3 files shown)


enricocupellini:
I think the way to use the IVMultiSliderControls with this constructor is subclassing it and implement "OnMouseDown", "OnMouseDrag", "OnMouseUp" methods.
Anyway is it ok to point out usage strategies on docs?

olilarkin:
hi enrico,

thanks for this. I feel you are going into a bit too much detail for the doxygen comments. They should describe the purpose of the method and the arguments, but they don't need to discuss about other methods. 

RE Multislider control - maybe the consrtuctor without an "action function" should have one. I need to think about it. That constructor is just for cases when your multislider is not linked to parameters

enricocupellini:
ok now I understand better the nature of trackable controls, no need to add actions, one can handle mouse changes using the other constructor, please refer to my last commit

Issue #33, opened 2018-11-11
Can IMidiQueue  be replaced with IPlugQueue?


olilarkin:

madronalabs:
I love tidying. I'll give it a try.

AlexHarker:
@madronalabs - have you taken a look at this at all?

I've just glanced at it but what I see is:

- IMIDIQueue doesn't appear to have any thread safety
- IPlugQueue is a fixed size queue, whereas IMIDIQueue can grow and shrink freely.
- IMIDIQueue uses a system of moving memory around that seems inefficient and unnecessary - it looks like it never loops and instead moves all the data around.

I pretty sure it can be done fairly easily with mostly gains, if we can live with the fixed size queue.

madronalabs:
All seems reasonable. I added the Resize method, see my PR

AlexHarker:
@madronalabs - probably Resize() should also clear, as it might resize downwards and that could invalidate the counters. Pulled it in, but it seems like that could be unsafe right now.

AlexHarker:
I have what I think is a rough version working here:

https://github.com/iPlug2/iPlug2/tree/IMIDIQueue-removal

this is ready for testing, but I have a couple of things to go through with Oli, plus a need to fix the counter issue on Resize() mentioned above.

olilarkin:
I am going to close this and keep IMidiQueue for now. I have too many things that depend on it. Will keep your branch elsewhere and may revisit it at some point

PR #265, opened 2019-07-19
Replace IMidiQueue with IPlugQueue

AlexHarker:
Designed to remove IMIDIQueue as per #33 

Could do with some midi testing. I've engineered it so that one can set a maximum number of MIDI message per second (currently 65536, but that should be agreed/reviewed). This should result in relatively small allocations, but also more-or-less consistent behaviour with different block sizes. My vote is to over-engineer the size so that the changes of running out of space are miniscule. The old system would grow indefinitely.


AlexHarker:
The PR also simplifies the IPlugQueue class (removing ElementsAvailable() and Peek() the later of which was unused and the former which was not needed) and makes the code using the code smaller also.

olilarkin:
seems in the IPlugInstrument example that when triggering notes via keys, note-off messages are not getting through, where as they were previously. Not yet tracked down why this is

AlexHarker:
OK - will check - is the AAX or VST3? Other formats *shouldn't* be affected as far as I'm aware.

olilarkin:
I tried this with the stand-alone app and VST3 in Reaper.  just to clarify IPlugInstrument  has some code for interpreting keystrokes and generating MIDI messages for some reason the note off messages are no longer working with these queue changes

AlexHarker:
Cool - I'll review

AlexHarker:
Have you had a chance to check MIDI from an external source?

olilarkin:
the branch is archived here: https://github.com/olilarkin/iPlug2/tree/IMIDIQueue-removal

I am going to delete it on the main repo now, as i try to prune the branches

PR #268, opened 2019-07-19
Copyconstructors

AlexHarker:
This is an interim PR that addresses copy constructors in IGraphics (and partially in IPlug).

I pospose pulling this in first and then following up later with an IPlug specific PR to finish this work.


olilarkin:
merged in #21274b8d239d1b94686ca935d7acf9509e0debc5

PR #273, opened 2019-07-30
IPlug APP wrapper transport bar implementation

hor-net:
This PR adds a transport bar to the APP wrapper that allows to correctly set the data into the APP ITimeInfo structure and updates them every time the audio call back runs

It also produces MIDI clock and MIDI start and stop messages, it also allows the APP to be a slave for other MIDI devices that sends MIDI clock, start, stop and SPP messages

Adds some switches in config.h to enable the transport bar and manage its look


olilarkin:
this PR is mixing IPlug and IGraphics, which is contrary to the iPlug2 ethos. The app "wrapper" code should not have any IGraphics stuff in it, since IPlug can be used with different GUI frameworks. I may separate the MIDI clock stuff and work that in in a different way

hor-net:
So basically i should remove all the methods of IPlugAPP where I use GetUI ?

olilarkin:
yeah, any transport capabilities added to the APP wrapper should be GUI agnostic.

PR #274, opened 2019-08-06
Fix AU timeInfo.mPPQPos being invalid at 0.0

hor-net:
0.0 is a valid PPQ position and ITimeInfo initializes mPPQPos at -1 so if the host sends 0 we need to accept it


olilarkin:
addressed with 249e6d010644ea3ac7d46a1a688d8fca9bbe0788, thanks

PR #276, opened 2019-08-09
ReleaseMouseCapture called too early 

hor-net:
when displaying info boxes the ReleaseMouseCapture is called too early, this destroys the IControl that called the ShowMessageBox method causing an access violation later on


olilarkin:
It doesn't destroy an IControl. Just commenting calls like this is not useful for a PR and what you have done introduces other issues so closing this

PR #278, opened 2019-08-09
Invalid ppq pos set by AAX wrapper

hor-net:
Sometimes PT sends a negative or invalid number in the

mTransport->GetCurrentTickPosition(&ppqPos);

call that is used to set TimeInfo ppqPos this fix prevent this from happening


olilarkin:
addressed with 5f4ae36559085a853b846471137aef44f5f49647, thanks

PR #280, opened 2019-08-09
make IVMeterControl able to receive just Data packet

jonata:
This makes it easier to send just float numbers to the meter (no need to create fake sample chunks).


olilarkin:
Thanks, makes sense and was planning to do it

PR #286, opened 2019-08-19
ADSREnvelope callback to notify when released

enricocupellini:
Use case:

I want to trigger an envelope with `sustainEnabled = false`

to do that I have a 2 states button (play/stop). When the envelope run is over according to _attack_ and _decay_ values I want my play button turns to the "play state"

therefore I can write:

`mAMPEnv.SetEndReleaseFunc([&](){
          static_cast<IVButtonControl*>(GetUI()->GetControlWithTag(mFilterEnvelopeButton))->SetShape(EVShape::Triangle);
          isPlaying = false;
      });`

I didn't find another way to achieve this other than adding a new callback triggered at the end of a release state. If a better solution already exists please discard the pull request and let us know the correct strategy.

thanks

### DIFF
--- a/IPlug/Extras/ADSREnvelope.h
+++ b/IPlug/Extras/ADSREnvelope.h
@@ -58,6 +58,8 @@ private:
   bool mSustainEnabled = true; // when false env is AD only
   
   std::function<void()> mResetFunc = nullptr; // reset func
+    
+  std::function<void()> mEndReleaseFunc = nullptr; // end release func
 
 public:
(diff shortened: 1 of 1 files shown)


olilarkin:
I have added this in e9bb6d84edbe468b2f42c32dc095ec2aa518144c, because i thought it was a cool thing to have, however the way you are using it is BAD PRACTICE! The function will get called on the audio thread, and so it is very important that you don't try and access UI controls and so on there. Better to set a boolean variable for example, and then update controls in OnIdle() on the main thread.

PR #289, opened 2019-08-24
Multiline Tooltips

PhilipeZenz:
Enables multiline tooltips. Breaks the tooltip at a certain width to make longer tooltips well readable. You can also use \n in tooltip strings now.

Just for IGraphicsWin.cpp

Would be useful for mac as well I guess, unless the mac graphics already have multiline tooltips.


olilarkin:
mac multiline tooltips already work

PR #290, opened 2019-08-25
Enabled GrayOut for ITextControl

PhilipeZenz:
Allows to GrayOut ITextControls. Also affects ICaptionControls.


olilarkin:
added in 6ea538f618020f248fd676a532981d0c10b7b880

Issue #257, opened 2019-07-16
IGraphicsWeb binds dblclick and mousedown events


olilarkin:
ends up confusing icontrols with mousedowns received and dblclicks for the same events

olilarkin:
https://github.com/iPlug2/iPlug2/tree/graphics-web_dblclick

AlexHarker:
So - there are some important things to note here:

1 - double click on web means after the mouse up for the second click (not the same as on desktop)
2 - the only way to get a double click at the same point as on desktop is to emulate, which means system settings have no effect on double click length etc. (not ideal)
3 - other options are:
A - map a different kind of action to double clicking on web
B - accept that double clicks will also cause mousedown/up events before the double click, but prevent double clicks being sent to controls that treat double clicks like single clicks

None of this is ideal, but that's how browsers work - thoughts?

AlexHarker:
The branch we have implements 2 - is there an issue with pulling it in?

olilarkin:
closed by b3c76d7bf5e752873e2952a91c1d46b9b9d324d8

PR #294, opened 2019-09-02
Graphics web dblclick

AlexHarker:
Relates to #257 

I think this should feel/act roughly the same as desktop.

There was a secondary issue on Mac in that if you keep clicking fast you can a single mousedown followed by a string of double click messages. On windows you get alternating mouse down / double clicks. I've adapted the Mac code to behave like windows for consistency (which is also how the web code is written).

It would be easy enough to remove the Mac change, but this way all three platforms should behave the same.


olilarkin:
Web dblclick seems to now work well, but with e.g. IPlugControls if I double click a knob on mac, it doesn't reset to default. Reverting 4738b2ebdeab5e241fd1242ea2a43d5bfb4f70f6 fixes it. Happy to merge this without 4738b2ebdeab5e241fd1242ea2a43d5bfb4f70f6

AlexHarker:
Yes - I am an idiot - should have been modulo 2, not 1 (which would make all clicks single clicks) - updated.

AlexHarker:
OK - now I've got it right I think (subtract 1 then mod by 2 to alternate)

olilarkin:
addressed in b3c76d7bf5e752873e2952a91c1d46b9b9d324d8 and 8d66f76eec029ba7483f7da8c160c39669c3423f

PR #298, opened 2019-09-06
Path based experiments

pressplay123:
these are just minor issues that I encountered on my journey into path based drawing. I wanted to refactor IStrokeOptions, too, and make it a class with proper constructors etc. But this would break code in many places and I simply didn't have the time to do it.

### DIFF
--- a/IGraphics/Drawing/IGraphicsNanoVG.cpp
+++ b/IGraphics/Drawing/IGraphicsNanoVG.cpp
@@ -192,6 +192,8 @@ static inline void NanoVGSetBlendMode(NVGcontext* context, const IBlend* pBlend)
 
 static NVGpaint NanoVGPaint(NVGcontext* pContext, const IPattern& pattern, const IBlend* pBlend)
 {
+  assert(pattern.NStops() > 0);
+  
   double s[2], e[2];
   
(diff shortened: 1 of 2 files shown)


AlexHarker:
There is a rationale for not having IStrokeOptions as a class. 

Firstly it doesn't have any invariants (all the members can be set independently of one another). 
Secondly, you probably don't care about setting everything, so having a struct is more convenient.

If you wish to add constructors to the struct to make it easier to set in one go then I'd be open to that, but I wouldn't want to make the members private (and hence there is no point making it a class)

pressplay123:
Okay, I agree to what you are saying, but... Currently, although IStrokeOptions is a simple struct, in order to initialize it with some arbitrary DashOptions you need to create a DashOptions object first and call SetDash on it. I would prefer to be able to create IStrokeOptions on the fly by means of an initializer list.

AlexHarker:
Agreed that constructors with initialiser lists would be good for ease, but just saying that there is no need to make it a class with private members, which might break other people's code without any benefits.

pressplay123:
Actually, the only thing I would change after having a second thought, is to add two constructors for DashOptions: 1. one that allows to set its members and 2. default one which makes sure that existing code is not broken.

AlexHarker:
Sounds good - if you make a PR then I would merge that in!

Issue #318, opened 2019-09-30
Lice and AGG backends crash when app window is dragged from retina to non retina screen


olilarkin:
GetPlatformContext() returns bad context in IGraphicsXXX:EndFrame():

This may have appeared on macOS 10.14.

olilarkin:
fixed on master

PR #319, opened 2019-09-30
invalidate platform context when moving UI from one screen to another

olilarkin:
fixes crash #318


AlexHarker:
This makes sense to me and looks fine in itself, but I'm wondering why the call to GetContext() in EndFrame() for LIce/AGG is returning invalid - what's the sequence of calls around IGraphics::Draw() and viewDidChangeBackingProperties when moving between screens? It seems like there is some kind of call to Draw() happening before we get viewDidChangeBackingProperties even though the context has changed.

Don't have a way of testing this easily here.

olilarkin:
going to merge it now. If you want to do some deeper investigation at a later date feel free to re-open

PR #320, opened 2019-10-01
Fix AAX plugins path

jonata:
This fixes the AAX plugins path to the default path, avoiding an error on Xcode.

### DIFF
--- a/common-mac.xcconfig
+++ b/common-mac.xcconfig
@@ -147,7 +147,6 @@ AUv3_FRAMEWORKS = -framework AudioToolbox -framework AVFoundation -framework Cor
 VST2_PATH = $(HOME)/Library/Audio/Plug-Ins/VST
 VST3_PATH = $(HOME)/Library/Audio/Plug-Ins/VST3
 AU_PATH = $(HOME)/Library/Audio/Plug-Ins/Components
-AAX_PATH = /Applications/Avid/ProTools_3PDev/Plug-Ins
+AAX_PATH = /Library/Application Support/Avid/Audio/Plug-Ins
 APP_PATH = $(HOME)/Applications
 REAPER_EXT_PATH = $(HOME)/Library/Application Support/REAPER/UserPlugins
-


olilarkin:
I have the latest PT Dev build available and it finds plugins in /Applications/Avid/ProTools_3PDev/Plug-Ins. And doesn't find them in the /Library/Application Support/Avid/Audio/Plug-Ins path 
@earlevel do you know why this might be... i can't remember changing a setting

earlevel:
I don’t know...I have a couple of dev builds on my Mac and they both find plugins in /Lib...it’s been since TDM days since 3PDev for me (with a long gap between), but the haven’t downloaded a recent dev build. And I’m sure I didn’t do anything for it to support /Lib.

But I’m away from the computer this week...

TBProAudio:
> 
> 
> I have the latest PT Dev build available and it finds plugins in /Applications/Avid/ProTools_3PDev/Plug-Ins. And doesn't find them in the /Library/Application Support/Avid/Audio/Plug-Ins path
> @earlevel do you know why this might be... i can't remember changing a setting

AAX load from /Library/Application Support .... never worked for us. I think it is just a PT thing.

jonata:
> it finds plugins in /Applications/Avid/ProTools_3PDev/Plug-Ins.

Supposing the Avid/ProTools_3PDev/Plug-Ins exists.

The latest Pro Tools Developer available for mac is version 2019.6.0.291. The DMG just tell to drag and drop the "Pro Tools.app" inside "Aplications" , no ProTools_3PDev mention. Here the path ⁩/Library⁩/⁨Application Support⁩/Avid⁩/Audio⁩ works just fine. Also, wouldn't be $HOME/Library⁩/⁨Application Support⁩/Avid⁩/Audio⁩ an option?

olilarkin:
ReadMe.txt that comes with the PTDev build explains how it can be useful to create a "Plug-ins" folder next to the .app which will be scanned. I prefer this for isolating unsigned binaries. This of course should be documented, but i am not going to change in xcconfig for now

PR #322, opened 2019-10-02
Middle C is c4

djowel:
Added option to have middle-C as C4 for IParam InitPitch.


djowel:
The middleCisC4 is a function parameter in this case. Another option is to add a flag to the EFlags, but I'm not sure if you want that. Tell me which API you prefer.

PR #323, opened 2019-10-02
General iPlug2 enhancements

gvampgh:


olilarkin:
thanks for the PR. but please do PRs atomically, not grouping loads of different tiny changes. The images.md is old and there needs to be much more serious docs reworking, so not merging that. don't really see the need for IColor.set, just use constructor. commit ce3a193ba8008cfaa183ee22f24923c191387718 modifies IControl::GetDelegate to return the IGEditorDelegate, which is sensible

PR #331, opened 2019-10-05
Add ClampZero to method IColor

gvampgh:
Clamp Zero has better performance if it's sure that the value can't be bigger than 255. Used since years in own IPlug1 fork. Would be great, if you could merge it. However, this is still my git beginning, I still don't know how to discard the documentation change out of the pr, sorry.


olilarkin:
please do it atomically!

olilarkin:
for each thing you want to be considered for iPlug2, make a branch with a single commit, or a few commits related to the same feature. Then do a pull request on github for that branch. Otherwise it is a lot of work for us to consider your PR. Thanks!

PR #334, opened 2019-10-05
Add ClampZero method to IColor

gvampgh:
For performance reasons, ClampZero is faster than Clamp (if it's sure that the values are not bigger than 255). Used in my projects for years.

### DIFF
--- a/IGraphics/IGraphicsStructs.h
+++ b/IGraphics/IGraphicsStructs.h
@@ -183,7 +183,10 @@ struct IColor
   
   /** /todo */
   void Clamp() { A = Clip(A, 0, 255); R = Clip(R, 0, 255); Clip(G, 0, 255); B = Clip(B, 0, 255); }
-  
+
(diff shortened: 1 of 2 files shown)


AlexHarker:
I don't personally see a strong case for including this. It's very unlikely that you'd get a speed bottleneck using Clamp() given the costs of actually drawing compared to drawing commands, and if you need something to clamp in this way you could write it is as a function without needing to modify iplug.

I'm also not really clear on why you'd know that the numbers are 255 or lower, but they also might be negative, so it seems like something for a specific use case and not a general one.

gvampgh:
Well, I indeed figured out a bottleneck with performance measurements calling std::min and max very often in Draw methods of my big projects. Since is maybe no real big issue for others, accepting my prs would highly motivate me in doing this. :-) I think the change wouldn't be a bad thing.

olilarkin:
we are not including this, sorry

PR #335, opened 2019-10-07
Add SetPosition and SetSize

gvampgh:
This adds my in the last years often used methods to set the position and size of controls after creation.


olilarkin:
added in fee0463cc5943842d4ee4536c09aa14536c9090a

PR #337, opened 2019-10-08
Add multiplier to IColor

gvampgh:
I used this very often in the past to set gfx and control colors in the ui by a user value/control multiplier.


gvampgh:
Closed due to 2 weeks absence of any attention

PR #339, opened 2019-10-08
Lock around calls to process audio if PARAMS_MUTEX is defined

### LINKED ISSUE #336: Synchronization problems due to removed mutexes
Example code:

```
void MyPlug::ProcessBlock(sample** inputs, sample** outputs, int nFrames){
	
	// In IPlug 1: // Mutex is already locked for us.
	
	const int channelCount = NOutChansConnected();
	
	for (int i = 0; i < channelCount; i++) {
                double* input = inputs[i];
                double* output = outputs[i];
		for (int s = 0; s < nFrames; ++s, ++input, ++output) {
			double mod;
			if (i == 0 || GetFlagSet(Stereo)){
				mod = GetMod(mModPos[i], mCurve[i])
				if (i == 0)
					mModCache[s] = mod;
				mModPos[i]++;
			}
			else{
				mod = mModCache[s];
			}
			*output = *input * mod;
		}
	}
	
}


void MyPlug::OnParamChange(int paramIdx, EParamSource source, int sampleOffset)
{
	// In IPlug 1: IMutexLock lock(this);
	
	double prmVal = GetParam(paramIdx)->Value();
	
	switch (paramIdx) {
		case kMonoStereo:
			if (SetFlag(Stereo, prmVal))
				mModPos[1] = mModPos[0];
			return;
	}
}
```

I have a function GetMod(pos, curve) that returns the value of a modulation curve at a certain position. If the stereo flag is not set, the mod variable is only calculated for the first channel and simply copied for the second, so both channels depend on mModPos[0] and mCurve[0]. If the stereo flag is set, mod is calculated independently for each channel and may use a different curve.
In OnParamChange(), when stereo is enabled, I synchronize mModPos which isn't incremented for the right chanel in mono mode.
In IPlug 1, that worked well because mutexes ensured that OnParamChange() and ProcessBlock() aren't executed at the same time. Thus, if mCurve was the same for both channels, there was no difference between mono and stereo mode. In IPlug 2 though, mModPos may be synchronized while ProcessBlock() is running, so if this happens when the interators i and s are not 0, the channels are no longer in sync and mModPos is not the same at the same sample pos of a channel.

Since this problem is likely to occur with any plugin that tries to keep synchronization between channels, it would be cool if IPlug 2 readded the mutexes that I Plug 1 had, maybe as an optional feature that can be enabled via a #define.

I realise that in the IPlugEffect example, the samples of the channels are processed alternating what would limit the syncroniziation offset to one sample. I had already tried that I IPlug 1 but it proved to heavy for more complex processing, I guess in IPlug 2, there's no difference?. I also realise that in the IPlugEffect example, the params are read out at the beginning of ProcessBlock(). That's not easy to implement though if something (like syncing mModPos) shall only happen when a param is actually changed, and also a waste of performance when operations with the param values are complex.

AlexHarker:
This should resolve #336

### DIFF
--- a/IPlug/AAX/IPlugAAX.cpp
+++ b/IPlug/AAX/IPlugAAX.cpp
@@ -337,7 +337,9 @@ void IPlugAAX::RenderAudio(AAX_SIPlugRenderInfo* pRenderInfo)
       ProcessMidiMsg(msg);
     }
     
+    ENTER_PARAMS_MUTEX;
     ProcessBuffers(0.0f, numSamples);
+    LEAVE_PARAMS_MUTEX;
   }
   
   // Midi Out
--- a/IPlug/APP/IPlugAPP.cpp
+++ b/IPlug/APP/IPlugAPP.cpp
@@ -144,5 +144,7 @@ void IPlugAPP::AppProcess(double** inputs, double** outputs, int nFrames)
 
   //Do not handle Sysex messages here - SendSysexMsgFromUI overridden
 
+  ENTER_PARAMS_MUTEX;
   ProcessBuffers(0.0, GetBlockSize());
+  LEAVE_PARAMS_MUTEX;
 }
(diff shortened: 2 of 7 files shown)


AlexHarker:
changes applied in 57259bad0ce9bc6f7bf1586b53e7a9adc6f1971d

PR #341, opened 2019-10-10
Correct libpng and zlib projects

gvampgh:


olilarkin:
Yes Alex is correct. Unfortunately we can’t change the fact that WDL includes its own copies of third-party libraries

PR #344, opened 2019-10-12
IControl: assert when gfx isn't attached

gvampgh:
Coming from IPlug1 it easily happens, that AttachGraphics is called after attaching controls. This crashes the initialization of controls.

### DIFF
--- a/IGraphics/IControl.h
+++ b/IGraphics/IControl.h
@@ -388,6 +388,7 @@ public:
   {
     mDelegate = &dlg;
     mGraphics = dlg.GetUI();
+    assert(mGraphics && "Graphics not attached");
     OnInit();
     OnResize();
     OnRescale();


AlexHarker:
SetDelegate is only ever called by an existing IGraphics, so I don't see what situation this helps to debug.

gvampgh:
No, please check this, I ran into several crashes on win. Attach gfx after ctrl attach and it crashes, don't use LayoutFunc or an override. Again, this won't be a bad thing to accept.

AlexHarker:
I have already read the code and checked it before I posted my previous response.

1 - I'm afraid you haven't explained how this change sorts the problem you describe. SetDelegate is only called by an IGraphics and therefore it is not possible for the delegate's mGraphics to be null at that point - so as far as I can see the assert will *never* fail - therefore this commit doesn't do anything in practice.

2 - in iplug2 (and possibly plug 1) you should be able to attach graphics after attaching the controls - I have code that does this. If you rely on being attached to the plugin when your control is constructed this is a design issue and you need to use  IControl::OnInit() - this is already documented.

gvampgh:
Well, I explained it, but you did't confirm that you have checked it in practise. I would like to highly appreciate a discussion before you close a crash issue !

olilarkin:
@gvampgh IPlug2 is open source. You can make your own fork and put whatever changes you want in it. If you want an assert there, have it in your version.

gvampgh:
I already have it in my fork, i just would like to help others who run into the same problem.

olilarkin:
can you post some code to illustrate?

gvampgh:
Yes, I will do it later

gvampgh:
Commenting these lines in IPlugEffect example crashes the plug-in. The point is, that an AttachGraphics before control attach prevents the crash. That's why the assert would help. 

IPlugEffect::IPlugEffect(const InstanceInfo& info)
: Plugin(info, MakeConfig(kNumParams, kNumPrograms))
{
  GetParam(kGain)->InitDouble("Gain", 0., 0., 100.0, 0.01, "%");

//#if IPLUG_EDITOR // All UI methods and member variables should be within an IPLUG_EDITOR guard, should you want distributed UI
//  mMakeGraphicsFunc = [&]() {
    IGraphics *pGraphics = MakeGraphics(*this, PLUG_WIDTH, PLUG_HEIGHT, PLUG_FPS, 1.);
  //};
  //
  //mLayoutFunc = [&](IGraphics* pGraphics) {
    pGraphics->AttachCornerResizer(EUIResizerMode::Scale, false);
    pGraphics->AttachPanelBackground(COLOR_GRAY);
    pGraphics->LoadFont("Roboto-Regular", ROBOTO_FN);
    const IRECT b = pGraphics->GetBounds();
    pGraphics->AttachControl(new ITextControl(b.GetMidVPadded(50), "Hello iPlug 2!", IText(50)));
    pGraphics->AttachControl(new IVKnobControl(b.GetCentredInside(100).GetVShifted(-100), kGain));
//  };
//#endif
}

olilarkin:
Ok i see the problem... but i think we need to fix it by calling IControl::SetDelegate() in a place that works for the new way and the old way

gvampgh:
Yes, I would appreciate it

AlexHarker:
My assumption in 1 from reading the code was faulty, but as @olilarkin says there are better fixes. I'm quite happy to have discussions but they need to involve some level of detail - such as code - in order to make things clear and identify issues. The most direct way to do that is supply code.

IGraphics now takes a delegate for construction, it would be possible for the graphics to attach themselves, which would remove the need for the user to attach them manually. Currently that would assert/crash if they reattach them, so we'd need to figure out how to avoid this.

One option would be to make IGraphicsWin a friend of it's own delegate and then make AttachGraphics private. Thus calls to AttachGraphics wouldn't compile.

PR #360, opened 2019-10-18
Introduce OnUnserializeParamSet

gvampgh:
This change
- prevents crashes when loading States from older plug-in versions (fewer parameters)
- introduces OnUnserializeParamSet, a method I use to customize the restoration of parameter values

### DIFF
--- a/IPlug/IPlugPluginBase.cpp
+++ b/IPlug/IPlugPluginBase.cpp
@@ -129,7 +129,13 @@ int IPluginBase::UnserializeParams(const IByteChunk& chunk, int startPos)
     IParam* pParam = mParams.Get(i);
     double v = 0.0;
     pos = chunk.Get(&v, pos);
-    pParam->Set(v);
+    if (pos > -1)
+    {
+      if (OnUnserializeParamSet(i, v))
+      {
+        pParam->Set(v);
+      }
+    }
     Trace(TRACELOC, "%d %s %f", i, pParam->GetNameForHost(), pParam->Value());
   }
 
--- a/IPlug/IPlugPluginBase.h
+++ b/IPlug/IPlugPluginBase.h
@@ -124,7 +124,13 @@ public:
    * @param startPos The start position in the chunk where parameter values are stored
    * @return The new chunk position (endPos) */
   int UnserializeParams(const IByteChunk& chunk, int startPos);
-    
+
+  /** Override this method to prevent setting/overwriting parameters from unserializing
+   * @param paramidx The index of the parameter that would be set/overwritten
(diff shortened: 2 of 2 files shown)


AlexHarker:
I can see the usefulness of checking the position in the chunk before setting, but I'm not sure if we need to add a new method. Can you not use chunks and  override UnserializeState and SerializeState?

gvampgh:
UnserializeState is still overridden. I use this custom method since years, it comes in handy when prevent saved ui parameters to be overwritten (e.g. by user choice in my custom preset manager). Afaik it was (and is) best practise to still call UnserializeParams at the end (or in) the overridden method UnserializeState. For me it works very well this way and hasn't a performance impact.

AlexHarker:
This is not an issue of performance, nor of the length of time you've been using it. Of course it it works for you that is great, and exactly the sort of thing that would make sense on your on fork. Here we must think about what is likely to be of more general use, and I'm concerned about adding yet another method that makes it harder to understand and reason through for most users.

I'm still not clear on how you are using it, so a more detailed explanation of what you'd need to customise might be helpful. It is worth saying that a VST3 plugin that doesn't restore *all* of it's parameters will (I think) fail in terms of the VST3 SDK plugin tester, and I think that might also be the same for audio units, so if you are using it to decide whether to restore a parameter or not then there might be a conflict between your design and what the designers of the SDKs think.

According to the current docs you should call UnserializeParams, but we could just change that to make it clear that this is not essential if you wish to change the basic functionality, in which case you should provide your own code to do the equivalent of UnserializeParams.

TBProAudio:
> 
> 
> UnserializeState is still overridden. I use this custom method since years, it comes in handy when prevent saved ui parameters to be overwritten (e.g. by user choice in my custom preset manager). Afaik it was (and is) best practise to still call UnserializeParams at the end (or in) the overridden method UnserializeState. For me it works very well this way and hasn't a performance impact.

Keeping plug parameter compatibility across versions is a tricky thing:-)
We finally managed it by using XML based parameter set instead of IPLUG internal static double set.
XML parameter set is string based and can handle easily new or missing parameters.
In any case: I agree with Alex: just overloading SerializeParams/UnserializeParams is enough to handle this case.  If you like i could provide code demonstrating it...

gvampgh:
As said, I use the method to override restoring of some (ui/gfx) parameters by user choice in a preset system. Since this is an user choice, VST3 and AU (& auval) don't have a problem with it.  - I don't know, what I could explain better. This works very well and could have a benfit to others. Thanks to TBProAudio for XML idea, I stay with my small and running solution in my fork. Thanks, I close the pr.

PR #363, opened 2019-10-18
Add double click gesture to resizer

gvampgh:
Adds double click gesture -> reset to default dimensions, is more intuitive than right mouse button.


AlexHarker:
Oli and I agree this is better, and I have applied it in d5dce0cb477ca8a06b96c829d052f2f2aae6bc6f 
I've also removed shift or right click from this functionality, so double-click is now the only way to do this.

gvampgh:
Great, thank you

PR #364, opened 2019-10-19
Prevent warnings from type conversion

gvampgh:


AlexHarker:
Thanks for this - this sort of thing also drives me mad.

Would you mind updating to use static_cast? I know that plug is very inconsistent in this regard, but for me static_cast is the correct/modern style and so if we're changing code we should ideally get that style in there too.

gvampgh:
I will do it the next days, or you are free to edit it in the merge.

AlexHarker:
Thanks. I'll merge once the change is made.

gvampgh:
@AlexHarker done

olilarkin:
thanks

PR #365, opened 2019-10-19
Fix Lice DrawRect

gvampgh:
Fixes DrawRect and ShowControlBounds for Lice, additionally restores IPlug 1 behaviour for this


AlexHarker:
Is there a reason for closing?

PR #371, opened 2019-10-22
Add missing semicolons

gvampgh:


AlexHarker:
The style currently is that the macros are used with semicolons. I noticed this recently when looking through the source.

@olilarkin - what do you think is best? I normally would define function-style macros with semicolons as in this PR and then not use semicolons when putting them in the source. I think having them defined here, and on the line of source code seems weird.

gvampgh:
Ok, I understand it, and why this hasn't produce syntax errors before, seems to be a matter of taste.

olilarkin:
which existing macros have semicolons on the end? i think it's better not to include semicolons

asedev:
There should not be semicolons at the end of macros. Might cause issues with macros used as parameters of functions. And it looks wrong when used within code.

gvampgh:
> 
> 
> There should not be semicolons at the end of macros. Might cause issues with macros used as parameters of functions. And it looks wrong when used within code.

Of course not all macros (e.g. the dozens of macros in the config) can have a semicolon at the end - this here is about coding macros.

AlexHarker:
Personally I'd love these not to be macros but templates, but that is a harder job....

gvampgh:
Yes, you are absolutely right, changing macros always causes recompilations. And e.g. for PARAMS_MUTEX it's weird, what happens, when it's defined in the config.

olilarkin:
closing, see #374

PR #375, opened 2019-10-24
Add clip facility to IText and lice implementation

gvampgh:
Atm, drawing custom texts inside of a control can't be bounded to the specified rectangle, it extends automatically and overwrites the boundaries. This implementation adds my (long time used) clipping member to IText and I added the Lice implementation. I can't add the other gfx implementations atm, maybe this could be easy for someone who is familar with these.


AlexHarker:
Thanks - Unfortunately, there is already a way to do this in the other graphics backends using PathClipRegion() that doesn't match this approach, so it doesn't make sense to pull this as is.

LICE is legacy so the best we could probably do is add the ability to control the clipping region, but I don't think this is worth putting time into, given that very few people will be using LICE on IPlug2.

PR #377, opened 2019-10-24
Fix SVG hole/solid detection with NanoVG

olilarkin:


olilarkin:
was #136

olilarkin:
Merging this. @AlexHarker if you think there is a better way, perhaps we can sort that at a later stage

PR #382, opened 2019-11-05
Fix VST3 crashes concerning resizing

gvampgh:
I ran into random crashes in the vst3 implementation - here windows & reaper - when restoring projects/states at program launch. After debugging it appears, that program/call method execution order from host side is sometimes "unconventional". These changes fixed it for me.

### DIFF
--- a/IPlug/VST3/IPlugVST3.cpp
+++ b/IPlug/VST3/IPlugVST3.cpp
@@ -271,8 +271,9 @@ bool IPlugVST3::EditorResizeFromDelegate(int viewWidth, int viewHeight)
   if (HasUI())
   {
     if (viewWidth != GetEditorWidth() || viewHeight != GetEditorHeight())
-      mView->resize(viewWidth, viewHeight);
-
+    {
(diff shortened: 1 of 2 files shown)


gvampgh:
Same crash appears at the second position in Ableton VST3 implementation and additionally in IPlug1.

AlexHarker:
Do you have steps or plugin code to reproduce? I haven't managed to test this yet, but whilst the fixes are clearly not harmful they point to something else going on which I'd like to investigate. 

As far as I can tell from reading (which I may not have got 100% correct) the UI should only ever be resized when it exists and thus there is a valid view. It's possible that resize is being called outside of the normal flow of things, but that would seem to be plugin specific, and not a bug in iplug2.

olilarkin:
i experienced a similar crash

olilarkin:
There is definitely something weird here with VST3PluginTestHost. Occasionaly when I drag the corner resizer i get a crash in IPlugVST3View::resize() because plugFrame is invalid.

Working on improving this on this branch, so closing this PR

https://github.com/iPlug2/iPlug2/tree/VST3/view

PR #385, opened 2019-11-14
Address a crash and warnings

v1ne:
Using Visual Studio 2019, I got this crash with the `IPlugControls` example and warnings. Address them.


olilarkin:
thanks, i have made most of these changes, except for the DirectSound crash. I want to spend some time looking at the standalone audio settings, will make an issue

PR #389, opened 2019-11-26
Detect keypresses of modifier keys (CTRL resp. CMD, ALT, SHIFT) with OnKeyDown() and OnKeyUp()

PhilipeZenz:
For Windows and MacOS, not for iOS and Linux.

OnKeyDown() and OnKeyUp() send an IKeyPress with key codes  kVK_SHIFT, kVK_CONTROL and kVK_MENU when these keys are pressed.

Additionally, IGraphics has a method GetModifierKeyPressed, that accepts kFSHIFT, kFCONTROL and kFALT. Could be translated to VK-Codes if you think it makes sense.

### DIFF
--- a/IGraphics/IGraphics.cpp
+++ b/IGraphics/IGraphics.cpp
@@ -1037,25 +1037,25 @@ bool IGraphics::OnKeyUp(float x, float y, const IKeyPress& key)
   return handled;
 }
 
-void IGraphics::OnModifierKeysChange(float x, float y, int flag){
+bool IGraphics::OnModifierKeysChange(float x, float y, int flag){
     int changedKey = flag - mModifierKeys;
+    mModifierKeys = flag;
     int code;
     switch(abs(changedKey)){
         case kFSHIFT:   code = kVK_SHIFT;   break;
         case kFCONTROL: code = kVK_CONTROL; break;
         case kFALT:     code = kVK_MENU;    break;
+        default:                            return false;
     }
     
     IKeyPress keyPress {"", code, static_cast<bool>(flag & kFSHIFT),
-                                    static_cast<bool>(flag & kFCONTROL),
-                                    static_cast<bool>(flag & kFALT)};
+                                  static_cast<bool>(flag & kFCONTROL),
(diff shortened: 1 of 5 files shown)


PhilipeZenz:
Do you plan to merge this, or do you have any concerns on this implementation? Just wondering if I can expect to see this in IPlug or not.

AlexHarker:
It may be that @olilarkin has thoughts about this. I've looked a couple of times and I have to say I don't understand exactly what the exact aim is here or the use case - it might be useful if you could explain those, as well as any new methods and their intended usage.

Thanks

olilarkin:
I'm not against adding this feature if we can get it to behave consistently on mac, win and web, but this PR won't merge with current tip and there are some stylistic things that need to be tweaked. I've started a rework of it here, which doesn't have GetModifierKeyPressed() since you can get that from the mouse event surely? Also it doesn't yet call OnKeyUp, and haven't looked at windows yet

https://github.com/iPlug2/iPlug2/tree/graphics/modifiers

PhilipeZenz:
With the GetModifierKeyPressed() function, I wanted to have sort of global access to the modifier keys because a mouse event is not called always when a modifier key changes (OnMouseOver() is only called when the mouse is moved) and because I cannot be sure that OnKeyDown() is redirected to the right control.
I'm using this to show additional information on the screen when a modifier key is pressed.

With our implementation, I could set up a function with  SetKeyHandlerFunc() to catch all key events, I guess. Will try this out when I find some time.

PhilipeZenz:
Found some time look at this again. 
For windows, the implementation is quite easy (see here: https://github.com/PhilipeZenz/iPlug2/tree/ModifierKeys_v2), and the keyHandlerFunc works perfectly.
To detect key down and up on MacOS properly, we'd need to store the latest state of the modifier flag anywhere and compare it to the current, I guess. Not sure where to store that value, maybe in a static variable?

tommitytom:
This stops `OnKeyDown` and `OnKeyUp` being called for modifier keys - which is required in my plugin

PR #390, opened 2019-12-04
Update IPlugVST2.cpp

TBProAudio:
Smart Bypass: this is a fix for an old iPlug1 issue. Some VST2 hosts implement plug bypass and expect the plugin to deal properly with PDC. Otherwise tracks get out of sync. So this fix implements VST2 effSetBypass
VST2 PDC Handling: fixes missing PDC reporting to VST2 host
nFrames > GetBlockSize() fix: There DAWs around (e.g. Resolve Arena) which set nFrames higher than blocksize


TBProAudio:
Closed

PR #391, opened 2019-12-05
Adding more instrumentation, adding vblank thread to notify main thr…

broms95:
Still new to this github stuff.  Adding code to handle vblanks in windows.  Uses gdi.lib only and will work with NanoVG and Skia.  Fixing issue with ValidateRect --- it really needs to be called in WM_PAINT for code paths where BeginPaint EndPaint are not called.

I went a little overkill on instrumentation and timing trying to understand vblank ---> post message to ui thread ---> WM_USER + 1 ----> WM_PAINT.

Not handling dirty rects properly.


olilarkin:
merging this work in manually, see #392

PR #399, opened 2019-12-28
Make GetConfigIO() return a const pointer (and associated const-corre…

AlexHarker:
…ctness)

Testing this for CI.


AlexHarker:
Applied as 42b48fac53404e31d3ff3f757a8d6e1c0b86b9a9

PR #419, opened 2020-01-16
do not extend existing path when calling PathArc (skia)

pressplay123:
Calling PathArc twice before calling PathStroke would result in a connecting line between the curves, which is not intended I believe. Haven't checked the other drawing backends, however.


pressplay123:
I just noticed this would break the behavior of `PathRoundRect` which would have to be adapted by adding a few manual calls to `PathLineTo`. This is probably also the case with `IVGroupControl::DrawWidget` whatever that does. And I know now that Cairo behaves identically, so it would have to be changed there, as well (and all other drawing engines I assume). I could add the missing parts if you like. Thumbs up or down?

pressplay123:
Another question: Would it have been better to open an issue first instead of a PR?

olilarkin:
yes please, make an issue and describe the problem in as much detail as possible, thanks!

AlexHarker:
closing for the reasons given in #431 - if necessary we can continue discussion there

PR #429, opened 2020-01-23
Added embossed style, embossed pressable rects

drakfluga:
This is an old piece of code I wrote about a year ago, updated and cleaned up for the latest iPlug2.

It adds an emboss option to IVStyle and (for now) draws embossed rectangle buttons. I plan to add the style to other drawing routines as well.


olilarkin:
I like this emboss option, but it breaks some layout of other shadows, not going to merge this PR as is but will think about how to add this feature in a clean way. thanks

olilarkin:
added emboss option... todo ellipse

drakfluga:
Thanks for including this! :)

Issue #117, opened 2019-01-27
Editor State recall between plugin loads


AlexHarker:
Currently the plugin editor state is valid only during the lifetime of the plug-in and is not reloaded. VST3 has a infrastructure to support recall, but other APIs are less obvious. This needs discussion.

AlexHarker:
@olilarkin - so currently there are hooks to serialize and unserialise editor state (which you can override for custom behaviour - by default just stores GUI size/scale). There's also a size/scale pair of methods which you can draw on to save that (as with (Un)SerializeState and parameters).

Do we think any of the formats should be automatically saving GUI state between sessions (obviously VST3 has a specific place to do this, but I'm unaware of anything like that for other formats)? This issue is open for that reason, but if we think it is the programmer's job to implement this then there is nothing to do. 

It could possibly also be provided as an option, although if it is happening through (Un)SerializeState (where I do it in the SRM plugs) the programmer is likely to override anyway, so that might not be a great idea. The difference for VST3 stuff makes this slightly annoying...

rexmartin158:
> 
> ... by default just stores GUI size/scale 

As I posted #773 it is not doing this by default.   When duplicating a plugin or saving a project in a DAW everything gets copied to the new plugin (all parameters, control positions, preset selected, etc.) _**except**_ for the UI size.  It's excellent - and highly desired behavior - that it's saving/copying all the control settings but what is required to get the host to save the UI size?

>
> 
> Do we think any of the formats should be automatically saving GUI state between sessions? 

IMO, plugin UIs should re-open exactly as closed/saved in the project.  I use several commercial plugins in my studio that do this.  The DAW (or the plugin somehow?) saves and re-opens the UI exactly as sized and set when the project was closed. Can't think of any reason why a user would not want this behavior.

olilarkin:
You can add it easily

https://gist.github.com/olilarkin/02a0618cde48f80e22f329513c044e5c

I think that since we need to maintain compatibility with existing state, it's up to the programmer to add this functionality

rexmartin158:
> 
> 
> You can add it easily
> 
> https://gist.github.com/olilarkin/02a0618cde48f80e22f329513c044e5c
> 
> I think that since we need to maintain compatibility with existing state, it's up to the programmer to add this functionality

Great!  Will give that a go.  Thank you for the info/link.

rexmartin158:
> 
> 
> You can add it easily
> 
[...]

PR #433, opened 2020-01-26
Graphics/resize recall

AlexHarker:
This is a refactoring of editor state code (including size/scale) so that it is more like the rest of plug, simpler and recalls correctly when the GUI is reopened.

At present it provides hooks for recall between DAW sessions but that is not default behaviour (this needs discussion as per #117


olilarkin:
Although this works. With NanoVG, macOS I noticed that when re-opening a scaled window, the call here to IGraphics::Resize() calls a few things that assume the context is valid.

https://github.com/iPlug2/iPlug2/blob/graphics/resize-recall/IGraphics/IGraphicsEditorDelegate.cpp#L33

i.e. nanovg context is null as is FBO, so it prints to the console "could not init FBO"

Later the resizing is correct, after other calls to IGraphics::Resize(), since the IGraphics::mWidth variables etc have been updated. Not sure the best solution

AlexHarker:
I checked the DrawResize() which looked safe, but I didn't look elsewhere. There are a couple of options but I was bound by the fact that there is no mechanism for passing sizes/scaling on creation.

AlexHarker:
Fixed the code so that error isn't shown and did some renaming - hopefully ready to merge now

PR #435, opened 2020-01-27
Implemented removal of single controls

TobiasKozel:
Allows removing controls with or without destroying the IControl.
Added a OnDetached function which clears the delegate.
Added a zIndex like render priority


olilarkin:
thanks, been thinking about adding something like this, will take a look

TobiasKozel:
Cool, I fixed the style a bit. If there's anything else wrong just let me know

AlexHarker:
Thanks for this.

@olilarkin might be able to comment on whether he likes the z-sort style in general, but I've noticed a couple of things that would need addressing before it is merged.

1 - GetControl() does not take a paramIdx - it takes control index

2 - There's something about naming and method parameters to RemoveControl that I'm not sure about. I don't like the deletion argument. If you're calling by pointer you are probably the control and I'd opt for DetachControl() and not delete. If you're calling by tag etc. then you are probably the developer just looking to delete and the old RemoveControlWithTag() would work

3 - As far as I am concerned it isn't up to the developer whether to set all controls dirty or not - when you take a control out it might have the side effect of needing a redraw and it's better to assume this is the case and *always* draw correctly, than to offer what is probably a very small optimisation that a developer might get wrong, leading to an incorrect rednder - happy to discuss this one, as maybe you have a scenario where this is more meaningful, but I'm not sure what that would be.

TobiasKozel:
Thanks for having a look.
Changing RemoveControl(IControl*, bool pWantsDelete, bool pSetDirty) to DetachControl(IControl*) sounds good to me. It will remove the control from mControls, call OnDetach() and SetAllControlsDirty() but not destroy the IControl.

As for 3, I made redrawing optional since RemoveControl() was also called from RemoveControls() for each IControl. Setting all controls dirty for each control removed seemed excessive. But if that's not a problem, then this option doesn't need to be exposed.

TobiasKozel:
Anything else that needs to be done?

olilarkin:
thanks for this one...appreciate the code commenting etc, just need some time to look at it and work out if its the right way to do this (wrt z order stuff)

olilarkin:
will look at this soon i promise :-)

AlexHarker:
Was there a reason for closing this? @TobiasKozel ?

TobiasKozel:
@AlexHarker Another pr was already merged in which handles the removal of controls.

AlexHarker:
Thanks - didn't catch that yet - hope that deals with your use case - just getting back into looking at things here after some busyness with other things.

TobiasKozel:
The removal part, mostly.
I also added some zIndex sorting. But this doesn't seem to be an important feature.
And let's be honest, it's been 4 months. I doubt this would be merged in.

PR #438, opened 2020-01-29
Make Creating Factory Presets Easier

dmsplugins:
Fixed an issue where cancelling out of a "DumpPresets" operation caused the plugin and host (Reaper) to crash. This issue existed in all "DumpPresets" functions except DumpBankBlob().
Fixed DumpAllPresetsBlob() so that it no longer outputs the extra chunksize argument to MakePresetFromBlob(). Speaking of which..... DumpAllPresetsBlob() and DumpBankBlob() save the same text to a file. Is there a reason for both to exist?
 I also modified the "DumpPreset" functions that write one call to a "MakePreset" function to append a new call to an overwritten text file each time the the "DumpPreset" function is called and added an option to add an incrementer between "MakePreset" calls to facilitate the preset creation process. There are some additional functions added for the same purpose. Please see the documentation in IPlugPluginBase.h for more detail.
 I also have some controls that I would like to send in another pull request, but some of them depend on whether this pull request is accepted and whether or not you decide to keep both DumpAllPresetsBlob() and DumpBankBlob().


olilarkin:
Here is roughly what I will merge of this, but I will make a new IPlug chunks example based on your additions

https://github.com/iPlug2/iPlug2/tree/presetwork_merge

There are a couple of preset manager controls i am working on here, and i will have a look at your about control too

https://github.com/iPlug2/iPlug2/blob/master/IGraphics/Controls/IVPresetManagerControl.h

olilarkin:
as discussed on slack, taking a bit of this

PR #440, opened 2020-01-29
added Set opacity and get with opacity

hor-net:
This PR add two handy methods to set the alpha channel and get a new IColor with modified alpha channel.
this saves some time in coding instead of creating a new IColor, assigning the old one then modifying the alpha value.
I took the idea from the GetContrasted methods.


olilarkin:
https://github.com/iPlug2/iPlug2/commit/697f9010ae0de0d3387f2d6f91f2a876c559a8aa

Issue #441, opened 2020-01-30
Add GetHSL method for IColor


hor-net:
The IColor struct already provides a GetFromHSLA method to create a color from Hue, Saturation and Luminance, but if you want to create a new IColor based on and existing IColor HSL properties you cannot do it.

I think adding a method to get current IColor HSL can be useful in this regard.

olilarkin:
7a6222b8293e187516ec792e8a8df4ffdfe268b8

PR #442, opened 2020-01-30
added GetHSL method to IColor

hor-net:
This PR adds a GetHSL method to IColor so that the HSL components are returned, it addresses issue #441


olilarkin:
merged, with changes in 7a6222b8293e187516ec792e8a8df4ffdfe268b8

PR #444, opened 2020-01-30
removed round from GetDisplayForHost to make it behave like IParam::Int that does just a type cast

### LINKED ISSUE #443: IParam::GetDisplayForHost and IParam::Int return two different values for integer parameters
**Describe the bug**
IParam::GetDisplayForHost has a call to round for integer parameters, while IParam::Int does not so a value like 1.8 is returned as 1 and displayed as 2

**To Reproduce**
Create an int parameter, assign it to a draggable control that also display its value, you can easily see the issue

**Expected behaviour**
I expect both IParam::Int and IParam::GetDisplayForHost to return the same values when used with integer parameters

**IMPORTANT DETAILS**
* What plug-in format does it relate to - VST2, VST3, AUv2, AUv3, AAX, APP or ALL? ALL
* What platform does it relate to - Windows/macOS/iOS/Linux or ALL? ALL
* What IGRAPHICS_BACKEND does it relate to, IGRAPHICS_NANOVG, IGRAPHICS_SKIA or ALL? ALL

**Additional context**

hor-net:
Removed a round in IParam::GetDisplayForHost when displaying integer parameters to bring the behaviour in line with IParam::Int that does just a cast to integer

This fixes #443

### DIFF
--- a/IPlug/IPlugParameter.cpp
+++ b/IPlug/IPlugParameter.cpp
@@ -285,7 +285,7 @@ void IParam::GetDisplayForHost(double value, bool normalized, WDL_String& str, b
 
   if (mDisplayPrecision == 0)
   {
-    str.SetFormatted(MAX_PARAM_DISPLAY_LEN, "%d", static_cast<int>(round(displayValue)));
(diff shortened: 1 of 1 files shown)


AlexHarker:
Closing after discussion on #443

PR #448, opened 2020-02-03
VST2: add missing DAW latency reporting, fixes issue 447

TBProAudio:
Description:
If plugin creates internal latency (e.g. convo engine) it needs to report this to VST2 host, otherwise tracks with the plugin get out of sync.
Currently iPlug2/VST2 does not report any latency changes to DAW via mHostCallback.

To Reproduce:
add SetLatency to e.g. ProcessBlock, DAW does not recognize new latency, no PDC

Expected behaviour:
VST2 plugin should report latency to VST2 host in order to enable proper PDC

### DIFF
--- a/IPlug/VST2/IPlugVST2.cpp
+++ b/IPlug/VST2/IPlugVST2.cpp
@@ -222,6 +222,7 @@ void IPlugVST2::SetLatency(int samples)
 {
   mAEffect.initialDelay = samples;
   IPlugProcessor::SetLatency(samples);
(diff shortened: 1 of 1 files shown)


olilarkin:
Which DAW had problems with this? Reaper seems not to need it

TBProAudio:
Right, Reaper does it without. Sorry, cant remember which DAW, need to check.
In any case, this is a very old patch (even IPLUG1 missed this). And you can find mHostCallback(&mAEffect, audioMasterIOChanged, 0, 0, 0, 0.0f) in Juce as well...

olilarkin:
OK... going to merge fingers crossed

Issue #465, opened 2020-03-05
IVSlider ignores roundess parameter in IVStyle object


hor-net:
**Describe the bug**
IVSlider ignores the roundness parameter of IVStyle

**To Reproduce**
Create an IVSlider control and use a style object with roundness specified, the track of the slider is still drawn as a rectangle

**Expected behaviour**
I expect the slider's track to have rounded corners

**IMPORTANT DETAILS**
* What plug-in format does it relate to: ALL
* What platform does it relate to: ALL
* What IGRAPHICS_BACKEND does it relate to: ALL

olilarkin:
added in ee9b85cfac495eb9bde4447ab9afa9bd6fbbb322

PR #466, opened 2020-03-05
consider mStyle.roundness when drawing the slider's track

hor-net:
This PR fixes the issue highlighted in #465

### DIFF
--- a/IGraphics/Controls/IControls.cpp
+++ b/IGraphics/Controls/IControls.cpp
@@ -675,11 +675,11 @@ void IVSliderControl::Draw(IGraphics& g)
 
 void IVSliderControl::DrawTrack(IGraphics& g, const IRECT& filledArea)
 {
-  g.FillRect(GetColor(kSH), mTrackBounds, &mBlend);
-  g.FillRect(GetColor(kX1), filledArea, &mBlend);
+  g.FillRoundRect(GetColor(kSH), mTrackBounds, mStyle.roundness, &mBlend);
+  g.FillRoundRect(GetColor(kX1), filledArea, mStyle.roundness, &mBlend);
   
   if(mStyle.drawFrame)
-    g.DrawRect(GetColor(kFR), mTrackBounds, &mBlend, mStyle.frameThickness);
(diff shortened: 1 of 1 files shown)


olilarkin:
I want to add something like this, but with the proposed code wider track sizes can look weird

image

really need to clip the inner track to rounded rect path, something like this with Skia, but not yet sure how to do it with other graphics backends

```
#ifdef IGRAPHICS_SKIA
#include "IGraphicsSkia.h"
#endif

void IVSliderControl::DrawTrack(IGraphics& g, const IRECT& filledArea)
{
  const auto cr = GetRoundedCornerRadius(mTrackBounds);

#ifdef IGRAPHICS_SKIA
  SkCanvas* canvas = (SkCanvas*) g.GetDrawContext();
  canvas->save();
  SkPath trackPath;
  trackPath.addRoundRect(SkiaRect(mTrackBounds), cr, cr);
  canvas->clipPath(trackPath, true);
  canvas->drawRect(SkiaRect(filledArea), SkiaPaint(GetColor(kX1), &mBlend));
  canvas->restore();
#else
  g.FillRoundRect(GetColor(kSH), mTrackBounds, cr, &mBlend);
  g.FillRoundRect(GetColor(kX1), filledArea, cr, &mBlend);
#endif
```

hor-net:
How did you manage to get that effect? I've set the radius and the track width the same and it seems to look good... of course if you use a radius higher than the track width you get a weird looking slider...

Screenshot 2020-03-09 at 12 24 30

olilarkin:
mStyle.roundness is a fraction 0-1, i presume you are setting it to something like 5?

olilarkin:
are you drawing the filled area?

hor-net:
Yes you are right i should have normalized to 0..1 but honestly i don't think this is useful at all, i think it's far easier to use the real radius when designing the interface, but even if I normalize that, 1 means: "the same radius as the track width" and i should not be able to go higher than one, so i still don't understand how the problem is created...

olilarkin:
addressed in ee9b85cfac495eb9bde4447ab9afa9bd6fbbb322, thanks

Issue #468, opened 2020-03-06
Add alignment for IRECTs


hor-net:
It would be very useful to have a way to align an IRECT to another, horizontally and vertically. 

Example: 

I want two IRECTs next to each other to be vertically aligned in their middle, i could call a method on the IRECT to align passing the other one as reference and get back an IRECT whose coordinates are changed so that now it's middle aligned with the other one.
The same goes for bottom, top, center, left and right align.

olilarkin:
added in 32a0e5279bf9986eeddde9c1d6b0393823e8ad86

PR #469, opened 2020-03-06
Added methods to align horizontally and vertically an IRECT to another

hor-net:
This PR implements the alignment of IRECTs as proposed in #468


olilarkin:
added in 32a0e5279bf9986eeddde9c1d6b0393823e8ad86

PR #482, opened 2020-04-01
BREAKING: rename IParam:GetXForHost -> GetX

olilarkin:
Shorten some IParam method names that have a pointless "ForHost" in their names


AlexHarker:
suggest we make this change ASAP

olilarkin:
done in e9dd286119e2ee98f1b941eda3f26215d1504da8

PR #492, opened 2020-04-11
Fixed crash when a ICaptionControl is constructed with no param linked

baktery:
With this fix a caption control can be created without any param linked to it during its construction, otherwise it crashed.


olilarkin:
done in 3b1acf88b98820a9e79d59de0eb7d19ffbf0649e

PR #494, opened 2020-04-12
Release DBGMSG macro fix

baktery:
Without the fix, cases like this will not work as expected:

https://github.com/iPlug2/iPlug2/blob/3b1acf88b98820a9e79d59de0eb7d19ffbf0649e/IPlug/APP/IPlugAPP_dialog.cpp#L546


olilarkin:
merged in 787aca73f44b1c4f230b736e13bf85fcdf41fbfe

Issue #474, opened 2020-03-21
Windows resizing flicker fix


Youlean:
Adding this code to PlatformResize fixes the problem. (This was accepted in the past but removed in some update)

```
  RECT r = { 0, 0, WindowWidth() * GetScreenScale(), WindowHeight() * GetScreenScale() };
  InvalidateRect(mPlugWnd, &r, FALSE);
```

olilarkin:
Unfortunate causes a lot of problems for the vsync mode, which is default

Youlean:
Hm, I haven't noticed it. The previous commit had UpdateWindow(mPlugWnd); which may cause the problem since it will force redraw with this function call.

Calling just InvalidateRect(mPlugWnd, &r, FALSE); shouldn't cause the problem since it just marks the window dirty. 

I will see how this behaves and let you know if I see any problems.

olilarkin:
Do you have flicker problems with NanoVG or SKIA GL with IGRAPHICS_VSYNC defined?

Youlean:
Skia and VSync with this change don't have flicker so far.

olilarkin:
for info: https://github.com/iPlug2/iPlug2/pull/497

Youlean:
This seems to improve flicker as well: https://docs.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmflush

Just put it at the end of: IGraphicsWin::PlatformResize(bool parentHasResized)

Youlean:
This is with the debug mode here. Note that the white edge is not visible in release mode here.
![Resizing](https://user-images.githubusercontent.com/13148293/80870833-3b459900-8ca9-11ea-8c24-8bc2c83eb8b3.gif)

gee-ell:
I see the same white-edge flicker on Windows 10 with in-built Intel HD 630, even in release builds.  And it's much worse than yours.  Will need to look at that.

PR #497, opened 2020-04-13
Resize window flickering on windows fixed

baktery:
On windows when the plugin window is resized an unpleasant flickering is producted in the window viewport.

This PR fixes that issue.

Please, if you could test this PR with other Windows systems it would be great because I only tested it on my machine.

Windows 10 64bit, nvidia GPU. Both NanoVG and Skia. IGRAPHICS_VSYNC enabled and disabled.

### DIFF
--- a/IGraphics/IGraphics.h
+++ b/IGraphics/IGraphics.h
@@ -1633,7 +1633,6 @@ private:
   bool mEnableTooltips = false;
   bool mShowControlBounds = false;
   bool mShowAreaDrawn = false;
-  bool mResizingInProcess = false;
   bool mLayoutOnResize = false;
   bool mEnableMultiTouch = false;
   EUIResizerMode mGUISizeMode = EUIResizerMode::Scale;
@@ -1644,6 +1643,7 @@ private:
 protected:
   IGEditorDelegate* mDelegate;
   void* mPlatformContext = nullptr;
+  bool mResizingInProcess = false;
   bool mCursorHidden = false;
   bool mCursorLock = false;
   bool mTabletInput = false;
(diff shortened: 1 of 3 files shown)


olilarkin:
which host in particular was this with, and was it on high dpi screens and low-dpi?

olilarkin:
If you turn on the FPS display (ctrl + F in app/debug build) you will see that if you move a knob this affects the frame rate, with IGRAPHICS_VSYNC on or off.

It seems like this is not the correct solution to the problem, if indeed there is a problem

fps

baktery:
You're completely right, I forgot to check if it was in the resize state.

I've just uploaded a fix.

baktery:
https://github.com/baktery/iPlug2/commit/8e10cc6900973d0631e80fa013ef18a0a40b6885

baktery:
Without the fix this is what I see:

bloom

baktery:
More info about the issue:

It seems like the flickering only happens with Nvidia cards.

Without this fix while you resize the window the OnIdle function (the timer used to call it) continue working normally.

With this fix the timer window message is queued but the redraw is processed before it. As a result of it if you don't stop moving the mouse the OnIdle function will be not called. 

So, not adding the fix:
- Pro is that the controls are updated regulary by the OnIdle function,
- Con is that the viewport flicks on nVidia cards (like the previous gif).

Adding the fix:
- Pro is that the window flickering disappears, 
- Con is that controls are not update while resizing the window. Maybe other unknown issues could appear because for now only tested by me (Nvidia) and by @olilarkin (ATI)

baktery:
I tried @Youlean suggestion  https://github.com/iPlug2/iPlug2/issues/474 of calling InvalidateRect inside the PlatformResize function and in my machine (Nvidia) the "black" flickering but a new "white" flickering appears :(

Issue #499, opened 2020-04-16
Examples break on case-sensitive FS


cfga:
**Describe the bug**
Building of the provided examples breaks in a case-sensitive environment (can't import module from `/Sources` dir, `IPlugControls-au-Info.plist` could not be found)

**To Reproduce**
Build one of the examples form scratch in Xcode (I ran specifically the AUv2 target) on a Mac OS with a case-sensitive FS.

**Expected behaviour**
The project builds without errors.

**Screenshots**
After trying to build:

![image](https://user-images.githubusercontent.com/12377093/79516011-65148400-804a-11ea-8a0d-30295b374fee.png)

After fixing `/scripts` dir (making it `/Scripts`):

![image](https://user-images.githubusercontent.com/12377093/79516141-c6d4ee00-804a-11ea-9e3b-5288b0d83148.png)

**IMPORTANT DETAILS**
* Using Xcode, mac OS X Catalina, APFS (Case-sensitive)

olilarkin:
should be fixed in 06cd33a8ccdb8126a448d45c638aed473908a921

PR #500, opened 2020-04-16
Fix examples breaking on case-sensitive FS

cfga:
This PR attempts to fix the problem stated in the issue #499: the examples break on a case-sensitive FS.

This is my first contribution attempt and I am just starting to learn and use IPlug2 :) Might have missed something (some other case-dependent stuff). This fix eliminates the problem for me.


olilarkin:
Thanks very much for this. I have merged your changes, slightly cleaned up in 06cd33a8ccdb8126a448d45c638aed473908a921

hope i didn't miss anything

cfga:
@olilarkin thanks :)

PR #502, opened 2020-04-18
Add flexible IVRadioButton area width

gvampgh:
The gap between the button and text seems to be very high, 1.5x is a good value (to me). However, I have implemented a new value in the constructor with the old default value 3x.


olilarkin:
it might be better just to have a pixel offset between button and label. 

thinking about it...

gvampgh:
I think your solution with the factor is allready great, I'm a fan of not having to set pixel sizes. :-)

PR #503, opened 2020-04-18
Add num cells to IRect GetGridCell

gvampgh:
The IRect cells are a great feature ! I added the possibility to get more than one cell as unified IRect. Useful e.g. for sub panel drawings and other stuff.


olilarkin:
I kind of like this, but I also think that doing the union manually is more flexible, and makes this method a bit less complicated. Here you can only make a union in the same direction as you specify

still thinking about it...

gvampgh:
I did this manually before and got for many panels some extra lines of code, that's why I was happy wih this branch here. :-)

PR #518, opened 2020-05-03
Native window maximize and resize supported on Apps

baktery:
For some plugins, when built as standalone app, it is interesting to enable the native window maximize button and the native window resize feature with the mouse (window borders).

IPlug2 already supported configuring the main app window dialog just modifying the IDD_DIALOG_MAIN resource, so enabling the window maximize button and setting the window borders as resizable was possible but it was not managed by the main window dialog function.

This PR adds the needed management on Windows. Native mouse resizing and native windows maximize button will work when the plugin is built as standalone app, a EUIResizerMode::Size type corner resizer has been attached to IGraphics and a maximize button and/or window border resize type has been enabled in the IDD_DIALOG_MAIN.

Native resizing lives together with corner resizer, so both resizing mechanisms are available at the same time.

It has been tested on Windows 10 with Skia and NanoVG,

### DIFF
--- a/IPlug/APP/IPlugAPP.cpp
+++ b/IPlug/APP/IPlugAPP.cpp
@@ -10,6 +10,7 @@
 
 #include "IPlugAPP.h"
 #include "IPlugAPP_host.h"
+#include "IGraphics.h"
 
 #if defined OS_MAC || defined OS_LINUX
 #include <IPlugSWELL.h>
@@ -35,9 +36,21 @@ IPlugAPP::IPlugAPP(const InstanceInfo& info, const Config& config)
   CreateTimer();
 }
 
+void IPlugAPP::WindowResize(int width, int height)
+{
+  igraphics::IGraphics* pGraphics = GetUI();
+
+  m_parentResized = true;
+  if(pGraphics && pGraphics->GetResizerMode() == igraphics::EUIResizerMode::Size)
+  {
+    pGraphics->Resize(width,height,pGraphics->GetDrawScale());
+  }
+  m_parentResized = false;
+}
+
 bool IPlugAPP::EditorResize(int viewWidth, int viewHeight)
 {
-  bool parentResized = false;
+  bool parentResized = m_parentResized;
     
   if (viewWidth != GetEditorWidth() || viewHeight != GetEditorHeight())
   {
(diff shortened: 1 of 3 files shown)


olilarkin:
iPlug2 should be usable with IGraphics or with another graphics library. This commit brings IGraphics into the app wrapper code, which is undesirable.

I have done some work already on this issue on this branch:

 https://github.com/olilarkin/iPlug2/tree/graphics/osresize

I would like to make the standalone app resizable using the OS window chrome, and fullscreen-able. This should be done in conjucntion with making the Web version fullscreenable... i.e. need to work out an API that suits multiple situations

https://github.com/olilarkin/iPlug2/tree/web/fullscreen

baktery:
> iPlug2 should be usable with IGraphics or with another graphics library. This commit brings IGraphics into the app wrapper code, which is undesirable.

I see, I didn't know it, it makes sense.

> I have done some work already on this issue on this branch:
> 
> https://github.com/olilarkin/iPlug2/tree/graphics/osresize

Cool, indeed, the first implementation I did was pretty similar to yours but it required to extend the IGraphics::Resize with and extra parameter to work properly, that's the reason why I decided to add it to the app wrapper side instead.

Do you have any plan to merge your branch with the main one?

I'm asking it because if that's possible I could test your branch and fix a issue I found when I implemented my approach (I didn't try your branch but it seems like it will fail when changing from maximized to default window size). I could make it working with the same limitation than the solution I proposed in the PR: For now to support resizing and maximizing (not fullscreen) on Windows for the "size mode" (ignore it when set to "scale mode" because this case is very complex, scale has to maintain both axis proportions and it will be very tricky to support it natively, even in that case maximize box makes no sense). As I said it will only require adding an extra parameter to the IGraphics::Resize() function.

If you prefer not to invest time on this for now because the final solution will be done with the OS window chrome approach you talked about (I don't understand it well but it's ok) then that's ok too.

baktery:
In case anyone needs the resize feature on Windows, I resolved it this way until an official solution exists:
[IPlugResizeHack.zip](https://github.com/iPlug2/iPlug2/files/4571394/IPlugResizeHack.zip)

Just include the IPlugResizeHack.h file and add this call inside your plugin OnIdle() function:

    ONIDLE_RESIZE_HACK();

This way no local IPlug branch is needed, using official master branch. It's a bit hacky but for now  the resize funcionality works on windows.

olilarkin:
I'm working on this here https://github.com/iPlug2/iPlug2/tree/graphics/onparentresize/Examples/IPlugResponsiveUI

currently borked on windows, but working on mac and web

PR #519, opened 2020-05-05
Correct label positions of shadowed radio buttons

gvampgh:
Labels of IV radio buttons are slightly offset, the shadow offset is not considered, this fixes it.

### DIFF
--- a/IGraphics/Controls/IControls.cpp
+++ b/IGraphics/Controls/IControls.cpp
@@ -478,7 +478,7 @@ void IVRadioButtonControl::DrawWidget(IGraphics& g)
     
     if (mTabLabels.Get(i))
     {
-      r = r.GetFromRight(r.W() - mButtonAreaWidth);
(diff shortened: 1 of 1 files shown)


olilarkin:
I don't agree. I think this change makes the buttons look less aligned with the text

gvampgh:
That's what I see in my own example. Left with fix and right without. Do you have an example where it looks less aligned ? Maybe there is missing an additional if condition.

grafik

gvampgh:
Ok, I revert this, it can't be the solution and I try to find an other way.

gvampgh:
I found the reason. IPlugControls uses EVShape::Ellipse, which works great. I use Rectangle, and this is in any way a little bit faulty.

gvampgh:
Rectangle draw uses GetAdjustedHandleBounds, it shifts the handle bounds by the shadow offset. Ellipse draw doesn't do this.

PR #521, opened 2020-05-05
Correct parameter value display

gvampgh:
At the moment all display values without a unit aren't centered below knobs (due to unnecessary extra space), this fixes it.


AlexHarker:
Makes sense to me - @olilarkin ?

PR #530, opened 2020-05-14
macOS: Switch to use CVDisplayLink instead of NSTimer, fix OpenGL on macOS

### LINKED ISSUE #26: Improve flicker/lag when UI first appears
Since enabling transient IGraphics contexts ( so that no textures et cetera are loaded onto the GPU  until plug-in user interface is  required), there is a noticeable lag seeing the UI,  at least with virtualCZ.  this wasn't an issue when the metal context was created before the UI was needed

olilarkin:
I believe this PR fixes #26, also fixes long standing issue of borked GL on macOS.

It completely removes the NSTimer, which maybe we should put back for the situation where you might want to have a lower FPS.


olilarkin:
damn, doesn't work well on catalina, closing for now

PR #543, opened 2020-05-24
Small docstring fixes in SynthVoice.h

mmontag:

thanks, we like PRs!

oli & alex


olilarkin:
thanks, feel free just to delete the placeholder text "Here are some instructions for a good PR" next time

PR #544, opened 2020-05-26
Allow control removal in IGraphicsLiveEdit

mmontag:
Add a menu with option to Remove when a control is right-clicked.
Also fix discrepancy where menu item "Add IVButtonControl" actually added an IVSliderControl.

Please review :) It works as desired but I'm not sure if address-of (&) is the right way to compare menus on L166 and L182.

### DIFF
--- a/IGraphics/IGraphicsLiveEdit.h
+++ b/IGraphics/IGraphicsLiveEdit.h
@@ -69,7 +69,11 @@ public:
       {
         mClickedOnControl = c;
         
-        if(GetHandleRect(mMouseDownRECT).Contains(x, y))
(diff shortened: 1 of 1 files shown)


olilarkin:
I've got various branches that improve the live edit control, but one of the major show stoppers is that I want to add a way of inspecting IControl properties that can work with custom controls. Work on this is on the controls/props branch.

anyhow, i have updated the live edit to support deleting and multiple selection, removing the add options for now

1f9d2efe7284a7f119ab6f6894478eb652371064

mmontag:
Cool 👍

Issue #540, opened 2020-05-20
Studio One macOS freeze using ICaptionControl's dropdown


hor-net:
**Describe the bug**
When using a plugin that includes the ICaptionControl and the control is linked an enum, the dropdown with the choices is shown. Showing the dropdown and choosing options works fine, when you quit Studio One the software hangs forever

**To Reproduce**
Build IPlugControls-macOS, no matter if VST, VST3 or AU, open it in studio one and click on any control that shows a dropdown, it doesn't matter if you select an option or just click outside the control do dismiss the dropdown. Quit Studio one, it doesn't matter if you save the project or not, Studio One will freeze.

**Expected behaviour**
I Expect he software to quit correctly as it does if no dropdown is ever shown/clicked

**IMPORTANT DETAILS**
* It relates to ALL plugin format
* macOS
* ALL graphics backend.

**Additional context**
I've tested it with both the custom popup menu control ad the system one and the issue happens in both cases. I'll try to drop a line to Studio One's team if the bug is confirmed by you.

olilarkin:
confirmed. there is a std::bad_cast in the stack trace

PR #546, opened 2020-06-02
added missing EndInformHostOfParamChangeFromUI on dropdown selection

hor-net:
Fixes issue #540 adding the missing call to EndInformHostOfParamChangeFromUI when an option is chosen from a dropdown or the dropdown is dismissed.

The BeginInformHostOfParamChangeFromUI is called on IGraphics::OnMouseDown and the EndInformHostOfParamChangeFromUI is never called for a dropdown since it's linked to the OnMouseOut event of the control that is never triggered for an ICaptionControl with a dropdown.


olilarkin:
thanks, i added this code in a slightly different place to make it get called on ios

merged in 4d7e86e

PR #550, opened 2020-06-08
fix Windows graphics-tablet mouse event detection

gee-ell:
As discussed in the forum (https://iplug2.discourse.group/t/windows-tablets-not-detected-correctly/), the original code does not detect my Wacom graphics tablet correctly on Win10 Pro x64 (pen mouse events do not set c_MOUSEEVENTF_FROMTOUCH signature).  The extended detection code that works correctly is adapted  from libSDL (http://hg.libsdl.org/SDL/rev/8254c364ec4a), these are their comments:

* Mouse data (ignoring synthetic mouse events generated for touchscreens) */ 
/* Versions below Vista will set the low 7 bits to the Mouse ID and don't use bit 7:
   Check bits 8-32 for the signature (which will indicate a Tablet PC Pen or Touch Device).
   Only check bit 7 when Vista and up(Cleared=Pen, Set=Touch(which we need to filter out)),
   when the signature is set. The Mouse ID will be zero for an actual mouse. */

### DIFF
--- a/IGraphics/Platforms/IGraphicsWin.cpp
+++ b/IGraphics/Platforms/IGraphicsWin.cpp
@@ -167,7 +167,8 @@ void IGraphicsWin::CheckTabletInput(UINT msg)
     const LONG_PTR c_MOUSEEVENTF_FROMTOUCH = 0xFF515700;
     
     LONG_PTR extraInfo = GetMessageExtraInfo();
-    SetTabletInput(((extraInfo & c_SIGNATURE_MASK) == c_MOUSEEVENTF_FROMTOUCH));
+    bool touchOrPen = !(((extraInfo & 0x7F) && (IsWindowsVistaOrGreater()? (extraInfo & 0x80) : 1)) || ((extraInfo & c_SIGNATURE_MASK) == c_MOUSEEVENTF_FROMTOUCH));
+    SetTabletInput(touchOrPen);
     mCursorLock &= !mTabletInput;
   }
 }


olilarkin:
I hooked up my Wacom tablet to my windows machine and I can't see any problems with the current behaviour. Could you describe the problem a bit more? I have my tablet (intuos 4)  in mouse mode not pen mode.

gee-ell:
Ah, I thought you had emailed me personally and responded via email a few days ago, didn't realise your Q originated here.

(I have the same sablet) I figured out the exact cause:

The Wacom driver control panel has an option to use the 'Windows Ink' framework, or not.  Many of us disable that because it causes those annoying splash-style click animations as well as other issues.

When Windows Ink is enabled, pen events do set the c_MOUSEEVENTF_FROMTOUCH signature.  But when it's disabled, it does not.  The extended check of the patch (taken from SDLlib) can separate pen from mouse events in that scenario, and they clearly added it for that.

So on Windows there are two ways to identify pen events, depending on if they got via the Windows Ink framework or not.

gee-ell:
this was never applied.  any update?

gee-ell:
friendly bump.  I've found no issues with the patch under various scenarios.

olilarkin:
sorry for the delay, i'll try it again soon and merge

olilarkin:
Ok, i'm just going to take your word for it :-)

olilarkin:
unfortunately it was problematic so I reverted this change (see https://github.com/iPlug2/iPlug2/issues/902)

PR #553, opened 2020-06-15
Allow ICornerResizerControl colour customisation (backwards compat.)

gee-ell:
Extends ICornerResizerControl with customisable colours for dragging, mouseover and idle, including the IGraphics::AttachCornerResizer() call.  Fully backwards compatible.

Especially useful for dark backgrounds where the current black colour is largely invisible.


olilarkin:
merged in 9daa201bae92e3d37ab09d826fb61b900140a831, thanks

Issue #470, opened 2020-03-06
Feature - Class Diagram?


check080:
I'm having a little bit of trouble reverse engineering the structure of a default plugin; i.e., which classes are held as instances of which classes, which classes are communicating, is IGraphics an instance contained within an IGEditorDelegate, how do the graphics editor delegate and the DSP "editor delegate" communicate, etc. My friends who have tried using IPlug generally have the same experience. The doxygen does a good job of showing inheritance, but if there were a class diagram for the default IPlugEffect that showed how all the classes are related to each other, including which ones are communicating, which one is keeping an instance of another class as a member, etc. then IPlug would be a much more approachable library while there are still not very many advanced code examples.

If I manage to figure out the class structure myself, I will attempt to draw a diagram, probably with [UMLet](https://www.umlet.com/).

olilarkin:
I've had several promises of help with the documentation, with no-one stepping up and delivering. 

I am not personally going to do this, but if you would like to be that kind soul who helps with this side of the project, that would be great and i'll happily explain things on slack.

I am going to rename this issue to be more generic and mark it as "need help" since i don't want to pollute our issue list with every single persons wants

QrchackOfficial:
There's support for that (class structure, at least in terms of inheritance) in Doxygen and it generates some diagrams. Might be worth a look as a quick improvement.

PR #560, opened 2020-06-21
Add documentation comments for (almost) all IRECT methods.

Bindernews:
I added documentation to all the IRECT methods except for Clank(), because I couldn't figure out what it's supposed to do. The most relevant issue I could find is #470, although this isn't a fix, it's just a contribution.


olilarkin:
Thanks!

PR #568, opened 2020-07-01
Add position tracking to IByteChunk for convenience.

Bindernews:
When creating an IByteChunk, it internally tracks the insert position. When reading from it, you have to manually track the read position, which is quite annoying. This PR adds variants of `Get`, `GetStr`, and `GetBytes`, as well as `Seek` and `Tell` to `IByteChunk` to internally track the read position. It also maintains compatibility with existing code.

I also added some documentation to a few of the functions while I was at it.

### DIFF
--- a/IPlug/IPlugStructs.h
+++ b/IPlug/IPlugStructs.h
@@ -111,7 +111,7 @@ struct IByteGetter
 class IByteChunk : private IByteGetter
 {
 public:
-  IByteChunk() {}
+  IByteChunk() : mPos(0) {}
   ~IByteChunk() {}
   
   /** This method is used in order to place the IPlug version number in the chunk when serialising data. In theory this is for backwards compatibility.
@@ -160,6 +160,17 @@ public:
   {
     return IByteGetter::GetBytes(mBytes.Get(), Size(), pBuf, size, startPos);
   }
+
+  /** Copy \c size bytes from the buffer into \c pBuf .
+   * @param pBuf Destination buffer
+   * @param size Number of bytes to copy
+   * @return Number of bytes read */
+  inline int GetBytes(void* pBuf, int size)
+  {
+    int len = GetBytes(pBuf, size, mPos);
+    mPos += len;
+    return len;
+  }
   
   /** /todo 
    * @tparam T 
@@ -181,6 +192,18 @@ public:
   {
     return GetBytes(pVal, sizeof(T), startPos);
   }
+
+  /** Copy a value out of the buffer at the current position and into the value and update the position
+   * @tparam T type of the variable to get
+   * @param pVal Pointer to the destination where the value will be stored
+   * @return int The number of bytes read from the buffer */
+  template <class T>
+  inline int Get(T* pVal)
+  {
+    int len = Get(pVal, mPos);
+    mPos += len;
+    return len;
+  }
   
   /** /todo 
    * @param str /todo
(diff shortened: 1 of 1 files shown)


mmontag:
Nice improvement! One suggestion: is there a more semantic verb than `Get` for indicating that an internal read pointer will be advanced?

Bindernews:
I tried to keep it consistent with the existing API. Currently you use `Put` to advance the write pointer, so I figured that just overloading `Get` would be the most consistent API. I'm not a huge fan of overloaded functions (except on rare occasions), but I did it for consistency sake.

mmontag:
+1 👍  ship it

AlexHarker:
This is a nice idea, but I'm not sure that the mixing of user tracked, and non-user tracked methods is great (@olilarkin may disagree). Mixing the calls might lead to more mistakes if you are altering code that is manually tracked but code autocomplete pulls up the wrong signature.

Another option would be to have a reader class that tracks it's own position but doesn't hold any data, which would be like a convenience class to just add the functionality you want.

Bindernews:
@olilarkin What's your take on this? Close it? Merge it? Move the code to another class? I'd like to get this resolved one way or another, just to close the PR.

olilarkin:
I agree that the IByteChunk shouldn't keep track of its own read position.

I've done some docs to refresh my memory of this stuff: https://github.com/iPlug2/iPlug2/commit/00acd3ecb6c0c795b98bcb1d491bf1c757720d73

Is it really very annoying? Would adding this helper class as @AlexHarker suggested be a satisfactory solution for everyone?

`
/** Helper class to maintain a read position whilst extracting data from an IByteChunk  */
class IByteChunkReader
{
public:
  IByteChunkReader(const IByteChunk& chunk, int startPos = 0)
  : mChunk(chunk)
  , mPos(startPos)
  {
  }
  
  /** Copy \c nBytesToCopy bytes from the managed IByteChunk into \c pBuf .
   * @param pBuf Destination buffer
   * @param nBytesToCopy Number of bytes to copy
   * @return Next read position in the IByteChunk */
  inline int GetBytes(void* pBuf, int nBytesToCopy)
  {
    mPos = mChunk.GetBytes(pBuf, nBytesToCopy, mPos);
    return mPos;
  }
  
  /** Copy arbitary typed data out of the managed IByteChunk at the current position and update the position
   * @tparam T type of the variable to get
   * @param pDst Pointer to the destination where the value will be stored
   * @return int Next read position in the IByteChunk */
  template <class T>
  inline int Get(T* pDst)
  {
    mPos = mChunk.Get(pDst, mPos);
    return mPos;
  }
  
  /** Retrieve a string from the managed IByteChunk and put it in \c str .
   * @param str Destination for the string
   * @return int Next read position in the IByteChunk */
  inline int GetStr(WDL_String& str)
  {
    mPos = mChunk.GetStr(str, mPos);
    return mPos;
  }
  
  /** Return the current position in the managed IByteChunk
   * @return The current position in the IByteChunk */
  inline int Tell() const
  {
    return mPos;
  }

  /** Set the current position in the managed IByteChunk.
   * @param pos The new IByteChunk position */
  inline void Seek(int pos)
  {
    mPos = pos;
  }

private:
  const IByteChunk& mChunk;
  int mPos;
};
`

Bindernews:
A helper class would be fine by me. I can change the PR to have the helper class.

PR #569, opened 2020-07-02
Add MIDI and polyphonic support in Faust target.

sletz:
Here is a general PR, updating the architecture files, but not the setup of the project itself. Not sure the code is completely clean, but I hope you get the idea!


olilarkin:
merged

Issue #571, opened 2020-07-04
Maintain a map of ctrlTags to controls


mmontag:
A few methods iterate over all `mGraphics` controls in order to find a control matching the provided `ctrlTag`:

- `IGEditorDelegate::SendControlValueFromDelegate`
- `IGEditorDelegate::SendControlMsgFromDelegate`
- `IGraphics::GetControlWithTag`

Improving this from linear to constant time by maintaining a map could yield a nice improvement:

- The number of controls can easily go into the hundreds for complicated plugins.
- These methods are called every frame if you are doing `IPlugAPIBase::TransmitData` in `OnIdle` to update the UI with messages from DSP code.

AlexHarker:
(@olilarkin -thouhgts?)

This seems like a nice/sensible idea, but it also occurs to me that GetControlWithTag currently returns a single tag, whereas the other two allow controls to share tags - is this a required functionality? If multiple controls with the same tag are required things change a bit - if not it seems weird that the loops in SendControlValueFromDelegate and SendControlMsgFromDelegate exist, because they could just call GetControlWithTag.

olilarkin:
I never intended controls to share tags. This seems like a good improvement, please could you guys take a glance at my PR and see what you think.

olilarkin:
done

PR #574, opened 2020-07-06
Use an unordered map to store ctrl tags

olilarkin:
see https://github.com/iPlug2/iPlug2/issues/571


AlexHarker:
This looks pretty sensible to me.

Issue #575, opened 2020-07-07
Plugin window incorrectly sized on windows when screen scaling set to 200%


hor-net:
**Describe the bug**
When you instantiate a plugin on a windows install that has the screen scaling set to 200% the ui drawing is correctly done at 2x size but the window size is not doubled so the user has to resize it manually to show the plugin content

**To Reproduce**
Just build one of the example and open it in Reaper setting windows' screen scaling to 200%

**Expected behaviour**
The plugin window should be large enough to contain the whole GUI

**Screenshots**
![VirtualBox_windows 10 x64_07_07_2020_18_20_49](https://user-images.githubusercontent.com/704273/86811831-a4b59f80-c07e-11ea-8050-940f67f5f7df.png)

**IMPORTANT DETAILS**
* What plug-in format does it relate to - ALL
* What platform does it relate to - Windows
* What IGRAPHICS_BACKEND does it relate to - ALL

AlexHarker:
Might be related to/a duplicate of #565

olilarkin:
fixed with merge, not related to #565 which is standalone app specific

PR #576, opened 2020-07-07
Correctly sizes plugin window on windows and 200% screen scaling

hor-net:
Fixes issue #575 simply moving the setting of mDrawScale before sizing the plugin window in IGraphics::SetScreenScale


AlexHarker:
This is obviously an ordering bug, and I've confirmed by reading the code that the change is correct, and would affect windows only. Should be merged. @olilarkin - can I leave you to do that, just to make sure it gets done in the preferred way?

olilarkin:
just trying it now

PR #579, opened 2020-07-08
Current work with adding CMake support to iPlug2

Bindernews:
Currently this setup (should) work with VST2, VST3, and standalone apps on Windows and Mac. There is no support for iOS, web, AU, or AAX builds. Also the resource bundling situation needs improvement for VST3 and bundles in general on MacOS. Regardless, the scripts are usable and there's even an easy way to add Faust DSP scripts as dependencies which will be compiled as part of the build process. Hot-reload Faust would probably work on Mac, but I haven't tested it.

Here is the `CMakeLists.txt` file for my plugin NESting, which should be considered a WIP example of how to use my CMake setup.
https://github.com/Bindernews/NESting/blob/master/NESting/CMakeLists.txt


Bindernews:
I know this is currently a draft, but what would it take to make this a full PR? Here's my current feature list:

* Windows (Visual Studio generator only)
  * ~VST2~ DONE
  * ~VST3~ DONE
  * ~APP~ DONE
  * AAX
  * Faust targets (Faust source -> hpp as part of build) (previously worked, probably broken with newer Faust)
* MacOS
  * AUv2
  * AUv3
  * ~APP~ DONE
  * VST2 - Should work but I don't remember off the top of my head
  * ~VST3~ DONE
  * VST3 - Separate
  * AAX
  * Faust targets (see above)
* Linux
  * VST2
  * VST3
  * LV2 (once iPlug2 supports it)

olilarkin:
closing this PR for now, since I believe there is more work on it elsewhere

PR #583, opened 2020-07-14
Documenting IGraphics.h WIP

olilarkin:


AlexHarker:
OK - most of this looks good - a few things:

- there are a couple of references to LICE as examples - perhaps these would be better referring to a different back-end?
- GetDrawContext() is unlikely to return a raw frame buffer (as suggested) - it most likely returns an underlying drawing context, according to which backend is employed
- "Draw a layer to the main IGraphics context, rotating it" - - I'd prefer "with rotation" at the end
- UpdateLayer()  the description can be - "Implemented by a graphics backend to prepare for drawing to the layer at the top of the stack"
- ApplyShadowMask() - "Implemented by a graphics backend to apply a calculated shadow mask to a layer, according to the shadow settings specified
@param layer The layer to apply the shadow to
@param mask The mask of the shadow as raw bitmap data
@param shadow The shadow specification"
- path drawing stack - I'm not sure what this means - there is no path stack - there is a transform stack (and a layer stack), but there is only one current path (no stack), so path methods add paths to the current path and path transform methods modify the current transform, whilst transform store/restores work on the transform stack
- param clearStates on PathTransformReset - Selects whether the call also empties the transform stack

I can probably do most of the edits myself if desired - just wanted to note everything that needs doing and do any discussion first.

olilarkin:
changed and merged, thanks

PR #584, opened 2020-07-15
Improve IPlugOSC, add WIP example

olilarkin:
- rework the IPlugOSC code to be more like iPlug2 style, and less like the osciibot style. 
- Modify behaviour to allow inheriting Sender & Receiver
- Allow changing of IP/port after instantiation
- Add example


olilarkin:
merged

PR #591, opened 2020-07-27
Allow layers/APIBitmaps to be explicitly cacheable

AlexHarker:
This PR primarily addresses an issue with Skia on GL (and possibly metal - untested) in which bitmaps that are scaled using ScaleBitmap() do not cache correctly, as they are bound to the drawing context of the loading plug-in.

As ScaleBitmap() calls StartLayer and the CreateAPIBitmap (where the actual backing for the scaling is created) I've added options to these methods to request a cacheable backing. That would also mean a layer could be created with a (raster) backing suitable for sharing between plug-in instances.

The only downside of this approach is that NanoVG backings are *never* shareable between plug-ins, so it gives the impression that you can request a layer that you could share between instances, which is not the case (as NanoVG bitmaps are simply IDs of the GPU framebuffer which is context-dependent). 

If preferred I can add a protected variable to IGraphics that holds this parameter specifically for the purposes of ScaleBitmap() so that the base drawing class can see it, but it isn't exposed to plug-in developers.

### DIFF
--- a/IGraphics/Drawing/IGraphicsAGG.cpp
+++ b/IGraphics/Drawing/IGraphicsAGG.cpp
@@ -480,7 +480,7 @@ APIBitmap* IGraphicsAGG::LoadAPIBitmap(const char* fileNameOrResID, int scale, E
   return new APIBitmap();
 }
 
-APIBitmap* IGraphicsAGG::CreateAPIBitmap(int width, int height, int scale, double drawScale)
+APIBitmap* IGraphicsAGG::CreateAPIBitmap(int width, int height, int scale, double drawScale, bool cacheable)
 {
   return new Bitmap(CreatePixmap<PixelMapType>(width, height), scale, drawScale, true);
 }
--- a/IGraphics/Drawing/IGraphicsAGG.h
+++ b/IGraphics/Drawing/IGraphicsAGG.h
@@ -266,7 +266,7 @@ public:
 
 protected:
   APIBitmap* LoadAPIBitmap(const char* fileNameOrResID, int scale, EResourceLocation location, const char* ext) override;
-  APIBitmap* CreateAPIBitmap(int width, int height, int scale, double drawScale) override;
+  APIBitmap* CreateAPIBitmap(int width, int height, int scale, double drawScale, bool cacheable) override;
 
   bool LoadAPIFont(const char* fontID, const PlatformFontPtr& font) override;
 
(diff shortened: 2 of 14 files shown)


olilarkin:
It seems fine. Default arg to cache doesn't seem to pass down.

```
diff --git a/IGraphics/Drawing/IGraphicsLice.cpp b/IGraphics/Drawing/IGraphicsLice.cpp
index 43f1d32b8..760be01a5 100644
--- a/IGraphics/Drawing/IGraphicsLice.cpp
+++ b/IGraphics/Drawing/IGraphicsLice.cpp
@@ -659,7 +659,7 @@ void IGraphicsLice::NeedsClipping()
     const int w = static_cast<int>(std::round(alignedBounds.W() * GetBackingPixelScale()));
     const int h = static_cast<int>(std::round(alignedBounds.H() * GetBackingPixelScale()));

-    mClippingLayer = std::make_unique<ILayer>(CreateAPIBitmap(w, h, GetScreenScale(), GetDrawScale()), alignedBounds, nullptr, IRECT());
+    mClippingLayer = std::make_unique<ILayer>(CreateAPIBitmap(w, h, GetScreenScale(), GetDrawScale(), false), alignedBounds, nullptr, IRECT());

     // Copy background in case of addition

diff --git a/IGraphics/Drawing/IGraphicsNanoVG.cpp b/IGraphics/Drawing/IGraphicsNanoVG.cpp
index 3b9889163..153331e88 100644
--- a/IGraphics/Drawing/IGraphicsNanoVG.cpp
+++ b/IGraphics/Drawing/IGraphicsNanoVG.cpp
@@ -402,7 +402,7 @@ void IGraphicsNanoVG::ApplyShadowMask(ILayerPtr& layer, RawBitmapData& mask, con
     IRECT bounds(layer->Bounds());

     Bitmap maskRawBitmap(mVG, width, height, mask.Get(), pBitmap->GetScale(), pBitmap->GetDrawScale());
-    APIBitmap* shadowBitmap = CreateAPIBitmap(width, height, pBitmap->GetScale(), pBitmap->GetDrawScale());
+    APIBitmap* shadowBitmap = CreateAPIBitmap(width, height, pBitmap->GetScale(), pBitmap->GetDrawScale(), false);
     IBitmap tempLayerBitmap(shadowBitmap, 1, false);
     IBitmap maskBitmap(&maskRawBitmap, 1, false);
     ILayer shadowLayer(shadowBitmap, layer->Bounds(), nullptr, IRECT());
```

AlexHarker:
Ah - sorry - If the call is from the same class the default needs restating on the class - do you want me to do that?

olilarkin:
yeah and you can merge it

Issue #513, opened 2020-04-30
ITextEntryControl can accept mouse-drag messages for selection, but they are blocked


olilarkin:
e.g. https://github.com/iPlug2/iPlug2/blob/master/IGraphics/Platforms/IGraphicsWeb.cpp#L329

AlexHarker:
This is not specific to web - see the PR.

AlexHarker:
Can this now be closed?

PR #609, opened 2020-08-23
Sort mousing to do with web and text entry

AlexHarker:
This PR fixes three mousing issues:

1 - the lack of mouse drags reaching mTextEntryControl (#513)
2 - web button identification being done using == on a bitmask (which could fail if multiple buttons are down)
3 - web button identification failing for mouse ups as buttons will always be zero 

Fixes are:

1 - use a different filtering scheme for mouse drags
2 - use bit mask testing instead (a & b != 0)
3 - use the button field (not *buttons*) to determine the button causing mouseup events

### DIFF
--- a/IGraphics/Controls/ITextEntryControl.cpp
+++ b/IGraphics/Controls/ITextEntryControl.cpp
@@ -560,9 +560,4 @@ void ITextEntryControl::SetStr(const char* str)
 {
   mCharWidths.Resize(0, false);
   mEditString = StringConvert{}.from_bytes(std::string(str));
-
-  if (mEditState.select_start != mEditState.select_end)
-  {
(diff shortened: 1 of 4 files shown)


AlexHarker:
I added a commit to deal with the issues with selection on TextEntryControl - that should now be fixed, assuming that DeleteChars and InsertChars are only ever called from stb_textedit.

Might be worth discussing the use of button/buttons on web - not 100% that we'll always be getting the flags we want, although what is in this PR is better than before, because mouse will come with button info, which previously it didn't (and as the mouseup in TextEnrtyControl looks at which button is pressed that can make a difference).

olilarkin:
This is all looking good and working well

Issue #105, opened 2019-01-13
Linux support [help wanted]


olilarkin:
Reaper now runs on linux and WDL/Swell has GDK support

It shouldn't be hard to make the stand-alone app run with Linux and to then implement LV2 support.  there are two branches for this with the beginnings of the work,  but I could do with help from someone who uses Linux more regularly.

IGraphicsLinux  is the main thing that needs implementing,  possibly using SWELL GDK support

see project: https://github.com/iPlug2/iPlug2/projects

olilarkin:
moving to project

AZSlow3:
I am not sure where to put a comment on that...

Using linux branch I was able to make IPlugMinimal ProcessBlock looping and primary dialog showing (without menu), so APP mode. 

I know, that is not much. But the amount of modifications to make it work was not zero ;)

I just want to say I have plans to make at least app work as desired.

olilarkin:
awesome. after messing around with the makefile, i thought maybe using cmake for linux would be better. on this branch there is some work on that https://github.com/iPlug2/iPlug2/tree/project/cmake

AZSlow3:
cmake can be nice looking and cross-platform solution... when it is already known how everything should be build. I sometimes prefer strait Makefile changes instead of "what is the best way to achieve that with cmake...". Also cmake is not good to compile and cross-compile in one go (my own projects are Linux/Window and I always compile them from one Makefile to be sure nothing is broken after any change).

I will start with Makefiles till app/vst3 works under Linux (with GUI), then we will see how it goes.
For VST2 I have no signed license. I think the best way is to use clean written VST2 ABI header, so not using anything from SDK, including characters "vst" nor other trademarks. Since there is no way to officially use VST2 ABI, as long as licensed files are not used, Steinberg can not enforce anything. They are in Germany/EU, where interoperability precedents exists (attempts to forbid using something, even still officially available, for the purpose of interoperability have failed in curt).

olilarkin:
I don't think VST2 support on linux is important, since it is effectively a dead format. VST3 is actually supported by steinberg on linux, and I think would be a good option for getting iPlug plugins running on ELK OS https://www.mindmusiclabs.com/. APP & VST3 would be great. I would also like to support LV2 eventually.

AZSlow3:
My current region of interest is REAPER under Linux/Windows. So f.e. LV2 will have no value for me (till REAPER start supporting it). I am also not sure I will participate in ELK target.

[...]

PR #616, opened 2020-09-08
Linux VST2, VST3, and APP support

Bindernews:
References issue #105 
I've spent the last two days or so working on this. I managed to get VST2 and VST3 working in REAPER without *too* much trouble, GUI and all. It's a little laggy when resizing the GUI, but that could be because I'm using a VM.

The build system uses CMake for maximum compatibility, and will even copy resources into the appropriate folders for each target type. Note that all resources are currently dumped into their respective folder or just the root resources folder, relative paths are WIP.

Now the problem: standalone apps. For some reason the GUI doesn't resize correctly when you open the app. The main window and all internal variables are correct, but the view only displays the top-left 100x100 pixels (that being the default size). I'm not familiar enough with X11 to know what the issue is, so I was hoping some assistance there (@AZSlow3 ?).

I also don't have LV2 support implemented yet, but that experimental LV2 support shouldn't be too hard to add given there's already some code for it.

I also need some decisions on the CMake scripts. When naming variables and functions should I use `iplug` or `iplug2` (e.g. `iplug_configure_target` vs `iplug2_configure_target`. I'm leaning towards just `iplug` given that the C++ namespace is just `iplug`.

Also: I need to add support for other IGraphics backends on Linux before I'd consider the CMake builds ready. 

TL;DR: VST2 and VST3 working (NanoVG, GL2, GL3), need help with App (looks like embedded X11 window isn't resizing properly), and LV2 is NYI.


olilarkin:
Thanks for this,Really great to have someone looking at linux support - I will have a look soon

Bindernews:
Looks like the original linux branch had more things changed than I thought. I'll have to diff and see what changed so I can add the appropriate `#ifdef`s.

Bindernews:
I've discovered another bug when resizing a plugin GUI to smaller than its original size, I suspect it has something to do with the plugin requesting the wrong size or something like that. 

Long-term I'd recommend re-writing the standalone app to use XEmbed like all other hosts. This would unify a lot of the code and make things more consistent between app and plugins. That being said, I'm not familiar enough with X11 to make that work. For now I'm just trying to get feature parity with Windows and Mac, and if it's not pretty oh well.

Bindernews:
GUI scaling is mostly fixed. Apps are still buggy, but that's a problem with the app host, not with `IGraphicsLinux`.

sonata-chen:
@Bindernews 
I tried to load the **IPlugInstrument**  **VST3** plugin but it failed.

Error message I got:
```
BEGIN IPLUG CHANNEL IO PARSER --------------------------------------------------
Channel I/O #1 - 0-2
               - input bus count: 0, output bus count 1
               - channel count on output bus 1: 2
               - input channel count across all buses: 0, output channel count across all buses 2

1 I/O configs detected
Total # in chans: 0, Total # out chans: 2

END IPLUG CHANNEL IO PARSER --------------------------------------------------
IGraphics NanoVG @ 60 FPS
INFO: resized to 1024x669
Could not locate font Roboto-Regular.ttf
```

I think the key point is the last line `Could not locate font Roboto-Regular.ttf`.
Any ideas? thanks.

P.S. I didn't get such error when running in **APP** mode.

Bindernews:
Yep. There was a typo in where CMake was copying the resources into, should be fixed with the latest push.

Bindernews:
@sonata-chen ^

sonata-chen:
> Yep. There was a typo in where CMake was copying the resources into, should be fixed with the latest push.

@Bindernews 
Great! I can load VST3 plugin now.

I found another error when compiling the VST2 version.
Here is the [error message](https://gist.github.com/sonata-chen/fb0745e33dc07ac5f50bcbcb74a1708c).

Bindernews:
Oh yeah. In `aeffect.h` on line 39 change `__cdecl` to `__attribute__ ((__cdecl__))`. It's just something you need to change if you're compiling with GCC. @sonata-chen

Bindernews:
I fixed it by adding a define in CMake.

Bindernews:
Here's a status update for what I think still needs to be done. I'd say Linux support is "good enough for now" once the Linux section is done.

Part of the problem with Linux right now is that we're only able to get around 45-50 FPS because we're not getting the callbacks fast enough. One option is to create a secondary "GUI" thread and shovel events over to it, but I'd say making it so that GUIs are framerate-independent is a better idea. Some machines may not have a good GPU.

Better CMake support can come in another PR.

To Do (Linux):
- Implement support for Skia graphics in CMake
- Create pre-built Skia libs for Linux x64
- Add menu for APP on Linux

To Do (IGraphics):
- Implement delta-time and update IVLEDMeterControl and IVDisplayControl to use it
- May need to implement timestamps on messages to sync properly 
- These changes are required for if the GUI isn't running at 60 FPS

To Do (CMake):
- Implement LICE, AGG, Cairo, and Skia graphics for all platforms
- Fix VST2 and VST3 on MacOS
- Implement AUv2 and AUv3 on MacOS
- Implement AAX (not happening because I don't have an Avid developer account)

Bindernews:
@olilarkin After some more examination, I think the reason menus aren't showing up is because they're not implemented in SWELL, likely because SWELL uses GDK and X11 stuff, and menus aren't a thing until you get to either GTK or Qt.

That being said there are a few options:
* Reimplement parts of SWELL to use GTK & GDK to create the window, menus, etc.
* Try to draw the menu ourselves (probably with Cairo, LICE, or something like it)
* Create an IGraphics context for the menu area and use IGraphics
* Say "never mind" and drop menus for Linux. Instead add CLI arguments to replace menu options.
* Instead of IGraphics use Dear Imgui for the menu only

I think using GTK is the best option, with by far the best platform integration, but it will require a lot of work and definitely won't be done any time soon if I'm doing everything. CLI args would be pretty easy and "Linux-ish", and we can put the menu on the back-burner. Creating an IGraphics context for the menu area is a good idea... but I don't think IGraphics is "independent" enough from the rest of iPlug2 to make that work well.

Bindernews:
Also after looking at REAPER on Linux, it seems that REAPER implements its own menus (which is probably why they look so ugly).

olilarkin:
Surprised about what you say about swell. We already have the option to draw the menu ourselves with IGraphics "IPopupMenuControl", so I wouldn't worry too much about menus.

I haven't got time right now to get deep into this, but quite a few of the suggestions above seem a bit odd (Implement delta-time etc). I am surprised if we can't get a good 60 fps drawing.

olilarkin:
ah ok... UI is drawing at FPS but something is up with OnIdle()... ?

Bindernews:
The problem isn't that we can't get 60 FPS drawing, the problem is we can't do it on the main thread. We only get the OnIdle() callback every 22ms for VST2, and since there's no "main" GUI thread unless we create one, the only place to do drawing is in OnIdle.

olilarkin:
I don't know why the swell app menus aren't showing, they did before

Bindernews:
Okay, maybe I'm building something wrong. I put a gdb breakpoint for anything reading the menu data, and nothing triggered it.

Bindernews:
Okay, looks like they defined SWELL_LICE_GDI, which means I probably have to do so as well, meaning I probably need to get LICE building with APP.

Bindernews:
We have a WINNER folks! Menus are working with APP (also confirmed that REAPER is drawing its menus using LICE which is why they look so ugly). For Linux support that leaves Skia, CMake, and cleanup, then I'll consider the branch ready to merge.

AZSlow3:
Just a summary of all my findings when I was looking into it (I have no time last months and will have no time next months to do anything in that direction):
1. VST3 has no OnIdle nor timers in the audio processing part. By design timers should not be used there (so any time based operations should be implemented using audio blocks processing). VST3 SDK/examples have generic "timers", but they are implemented using GUI timers (so VSTGUI code) and obviously work when the GUI is open (VST3 GUI API part support timers).
2. REAPER does not have proper implementation for timers not file descriptor waiting (defined in VST3 GUI API Linux), it simply call "ready" using the same approach as OnIdle (probably from the same place in the code)
3. SWELL is not written to work in plug-ins (on Linux). It (implicitly) use glib/GDK main loop. If it does not exist (f.e. in not GDK based hosts), there is no clear approach what to do (notice Cockos plug-ins do not exist as independent VST2 for Linux, they are SWELL based).
4*. As can be seen in my issue for VSTGUI, VST3 Host example in official SDK is disaster (many bugs just in several lines of code). I guess the author had no idea what he/she was doing (definitively has not written any Linux main loop before). The whole scenario of VST3 GUI API for Linux is more or less "dirty hack".

Bindernews:
Yeah, I've come to most of the same conclusions. Thanks for checking in @AZSlow3 . I still think the best solution is to have a main loop GUI thread, and just shuffle events from callbacks into that thread.

Bindernews:
@olilarkin I'd say this is ready for review. We can build Skia, and things seem to be mostly working correctly, other than the known issue about the GUI not running at 60 FPS. I'd like to get some other people testing things before I call it totally ready, because I'm sure I haven't hit every edge case.

We can discuss the 60 FPS issue later.

olilarkin:
great! I might try building one of my projects with it

olilarkin:
Could the cmake folder go inside Scripts?

olilarkin:
merged into linux branch, where I will tweak a few things. Does this include everything from your CMake PR too? I have a lot of work todo when I merge project/new_installer, and i will try and do that at the same time as merging linux/cmake so that each project compiles

Bindernews:
Sorry for the radio silence, I got super-busy with personal stuff. I've just pushed fixes for CMake on MacOS, and I should be able to test CMake on Windows soon enough. At this point all the cmake work is merged into my linux branch, and I'm working on Linux and CMake simultaneously. I know that's not best practice, but it makes things notably easier.

olilarkin:
Not a problem, ￼ i'm considering the Linux branch the Linux and see Cmake branch. I think we'll ditch the other make files for Linux￼

Bindernews:
That was my plan as well. The makefiles are okay, but CMake is much more configurable, and the fewer build-systems we have to maintain the better. Personally I'd love to see CMake become the "official" way to build iPlug2, simply because then we only have one build system to maintain. That being said I can also understand why would want to keep the default VS and Xcode projects. They look much nicer.

PR #623, opened 2020-09-22
CMake, Platform abstraction etc.

wizzie2:
I know this is not the usual way PR are done, I apologize. And let me also apologize for the code formating. I tried to use 2 spaces indent, but it was really really difficult to read the code. Looks nice, but hard to read, especially in longer multi indented code. Again, I apologize.

This PR is not complete at the moment, and since I only have a PC and Visual Studio, I can't make the Mac or iOS parts.
I've added CMake support with pre-compiled headers and an easy to use interface (that was the goal anyways). At the moment it only supports APP and VST3.

There is a structure for added for platform abstracted classes, and some some fixes for VST3 when UNICODE_OFF is defined. I've also added a basic math lib which is WIP like most things. Just wanted to provide some progress.


olilarkin:
thanks for the contribution,I think you're aware but it's impossible for us to consider big pull requests such as this. if you have some stand out features on your branch, please make individual PRs, if you would like us to consider them

olilarkin:
I'm closing this PR, but will keep an eye on your fork

wizzie2:
Oups, thought I commented earlier. But apparently only in my mind. The fork isn't really that big, it just looks that way due to the formatting and integration of cmake. Since it uses pre-compiled headers I had to remove a lot of includes in the source files. Cmake also provides a lot of definitions used by the source when compiling, which is why current implementation of CI won't compile it. That and hard coded #error for untested platform code that I can't implement at the moment.

btw, If there were a development branch or even dev_iplug_3(?) branch, I would have chosen that one instead of master. :)

PR #644, opened 2020-11-02
Changes for OpenGL 3 and Skia master

kcbanner:
This PR updates the Skia dependency to master, and fixes some problems I ran into with OpenGL 3.

Skia changes:
- `GrContext` changed to `GrDirectContext`
- The `svg` module moved
- Deprecated `SkMatrix` functions were removed
- There was a problem with the destruction order of the `SkSurface`. The OpenGL context was destroyed before the surface was deleted, and it would end up spin looping on `glGetError` in `GrGLGpu::clearErrorsAndCheckForOOM`. Since there was no context, it would return `GL_INVALID_OPERATION` and never exit the loop. Changed to destroy the surface/canvas/context in OnViewDestroyed.
- C++ standard changed to C++17, due to a mandatory RVO, see https://groups.google.com/g/skia-discuss/c/IAsnL2use7E/m/0Smg0vRPBgAJ

OpenGL 3 Changes;
- `GL_STENCIL_BITS` is undefined with the 3.3 core profile, changed to use `glGetFramebufferAttachmentParameteriv`
 
Windows changes
- Include the `wglCreateContextAttribsARB` extension
- Create a 3.3 core profile context if possible

I tested these changes on MSVC 2019, Skia master (4691b0f82bd00e799386a9d4327282d0a2c074ca), using IPlugEffect changed to use IGRAPHICS_SKIA and IGRAPHICS_GL3.

### DIFF
--- a/Dependencies/IGraphics/build-skia-win.sh
+++ b/Dependencies/IGraphics/build-skia-win.sh
@@ -57,7 +57,6 @@ if [ "$#" -eq 2 ]; then
   skia_enable_skparagraph = true
   skia_enable_sksl_interpreter = true
   skia_enable_svg = true
-  skia_enable_tools = false
   cc = "clang"
   cxx = "clang++"
   clang_win = "C:\Program Files\LLVM"
@@ -66,8 +65,8 @@ if [ "$#" -eq 2 ]; then
   if [ $CONFIG_STR = "Debug" ]; then
     echo 'extra_cflags = [ "/MTd" ]' >> ../../tmp/skia/$DIR_ARCH_STR/$CONFIG_STR/args.gn
 
-    echo 'is_debug = false' >> ../../tmp/skia/$DIR_ARCH_STR/$CONFIG_STR/args.gn
-    echo 'is_official_build = true' >> ../../tmp/skia/$DIR_ARCH_STR/$CONFIG_STR/args.gn
+    # tools currently fails to build with an SkPDF error
+    echo 'skia_enable_tools = false' >> ../../tmp/skia/$DIR_ARCH_STR/$CONFIG_STR/args.gn
 
     # disabled due to massive binaries
     # echo 'is_debug = true' >> ../../tmp/skia/$DIR_ARCH_STR/$CONFIG_STR/args.gn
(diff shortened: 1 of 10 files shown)


kcbanner:
Looks like the Skia builds are failing because the CI uses the pre-compiled versions. I'm going to update the skia dep to the latest chrome stable branch and verify these changes.

olilarkin:
yeah. I can rebuild the dependencies when your happy with the PR

kcbanner:
Updated for chrome/m87.

However, I discovered a bug in Skia when drawing 0-size dashed boxes. This can happen in live edit mode, if you click and release on the background, drawing a selection box that has overlapping corners.

I made a PR https://github.com/google/skia/pull/68 there for the change. This could also be fixed in iPlug by checking for this case before drawing. 

Is it preferable to add this change as a patch in skia.patch, or to make the changes on the iPlug side?

olilarkin:
we should probably clip that in iplug anyway

kcbanner:
Added IRECT::ZeroArea to cover this case, I think with that it's good to go regardless of if the Skia changes are merged.

AlexHarker:
Can I suggest that Empty() should be considered for updating, rather than adding ZeroArea() as another method?

We need to think about what situations we want to handle here (including possible floating point comparison issues and negative areas, where the orderings of the sides is reversed) - the code proposed here for ZeroArea() tells you when the rect has no area (regardless so position) OR when one set of parallel sides is swapped. Empty() simply tells you if all the values are zero.

@olilarkin - what are your thoughts on this? The comments on Empty() do make it clear what it does, but I'm not sure if this is the concept we want or not.

kcbanner:
To me it seemed like Empty() was more of a "null box" check, and not knowing how user code currently uses it I didn't want to touch that - maybe `ZeroArea` isn't necessary and code could know to just do `Area() <= 0.f`, but it does save the multiply.

olilarkin:
will get this working on the skiachrome87 branch

olilarkin:
This is now merged, apart from the dash/live edit issue. I can't reproduce. Could you make another issue about that.

DEPS updated to skia/chrome87

PR #661, opened 2020-12-08
fixed bug: ISVGControl prevents termination of the plugin (Skia/OGL/Win)

asedev:
The ILayerPtr has to be released in the destructor of the ISVGControl, otherwise, the Skia destructor fails.

### DIFF
--- a/IGraphics/IControl.h
+++ b/IGraphics/IControl.h
@@ -1855,7 +1855,10 @@ public:
   , mUseLayer(useLayer)
   {}
 
-  virtual ~ISVGControl() {}
+  ~ISVGControl()
+  {
+    mLayer.release();
+  }
 
   void Draw(IGraphics& g) override
   {


AlexHarker:
Thanks for spotting this issue. However, if we have to release layers in destructors it implies that the layers aren't being handled properly in relation to GL - this might be similar to the issues in NanoVG and it would be best to try and deal with this more automagically if we could. Our NanoVG has a way of doing something like this that I wrote, but I can't remember exactly how it works - we might need to keep a record of the layers or backing textures somewhere.

asedev:
I can see your point. But on the one hand side, since the layers are created in the ISVGControl it is just reasonable to "release" them there. And on the other hand side, this (or another) fix is needed since Skia/OpenGL using SVGs is "broken" currently.
I also tried to "pop()" the actually layers managed in IGraphics but this didn't solve it.

olilarkin:
I believe this is fixed by this commit, making sure controls are deleted before the GL context:

https://github.com/iPlug2/iPlug2/commit/1ffd729d1908ef11bb604d9a5b825d09c0b671ab

PR #667, opened 2020-12-19
fix GetControlWithTag

pressplay123:
The `at()` function of `std::unordered_map` does not return a nullptr if the element does not exist. It throws. I am very sure the intended way was to use `find()` and check the result. That's what the API suggests.


olilarkin:
thanks!

PR #680, opened 2021-01-13
IGraphics::mScreenScale becomes a float 

olilarkin:


olilarkin:
This fixes scaling on windows, when the screen scale is set to e.g. 125% or 150%

AlexHarker:
There are some details to consider in terms of bitmaps here - maybe that's been done, but they were the main reason screen scale was an int before.

olilarkin:
Yeah, I decided to just use 2x bitmaps if scaling is > 1x. I've tested all the platforms except web and seems to be working. Layers are the thing I am not too sure about

AlexHarker:
Yes - layers might be an issue. I'll try to take a look soon. It depends how pixel alignment is calculated and whether API bitmaps are allowed to have float scaling values.

olilarkin:
VST3 view is the wrong size initially on 200% screen with this for some reason

AlexHarker:
Can you explain what the count++ line is for - that looks very hacky and also I don't think it'll work even to do what you want as the static will be static to the method (and hence class), and not the instance of the GUI/plugin, which I presume is what you want.

olilarkin:
yes its a total hack that i am not going to merge, got pushed by accident.
When opening the VST3 in a DPI aware host (VST3PluginTestHost or  Cubase)
on screen with > 100% scaling, the window size is initially wrong. Dragging
the window to the 100% screen, window size is correct and back to the >
100% screen it is now correct. This hack was trying to work around that.
Might be easier to demo the problem with a video/chat

AlexHarker:
Cool - I assume you are referring to the VST3 DPI awareness? It would be good to look at that in a more detailed manner.

olilarkin:
hmm, reverted that and not seeing the problem any more. Maybe because I upgraded VST3TestHost

olilarkin:
@AlexHarker any chance you might be able to look at this soon? I'd like to merge it. Ignore reverted VST3 stuff, that stuff was a bug in cubase 10/old vst3 host, fixed in 11

PR #693, opened 2021-02-10
add quotes around b64 javascript string (otherwise it's a js error)

mazbox:
WebView callbacks don't work because the base64 strings in the EvaluateJavascript() function need to be surrounded in quotes (because they're strings).


olilarkin:
Hi Marek, going to merge and trust you are right cos' I don't have a way to test this right now. If you fancy improving that IPlugWebUI example to test all these things, be my guest 😉

mazbox:
This is only tested on mac vst3 and iOS auv3, not windows but I'm guessing it will have the same issue. Will have a think about revamping that demo. Also got some auv3 webview fixes on the way with regards to resizing, crucial for AUM

PR #704, opened 2021-03-14
set text and colors of IURLControl

pressplay123:
This is just a small extension of `IURLControl` which did not provide means to update colors (useful for theming). I used the same logic as the constructor where the original color is set implicitly through the `IText` argument and the other colors are set explicitly.

I would usually also add a `SetDirty` but I guess this is the responsibility of the caller?


pressplay123:
I would still like this to be merged.

olilarkin:
Added in https://github.com/iPlug2/iPlug2/commit/bdc6c02cf2ee327bddc5d43dcda82c5b4fdaf574, thanks !

Issue #712, opened 2021-03-25
Stepped controls not always working


pthvogt:
The stepped controls do not work correctly with step size < 1.0.
`OnMouseDrag()` and `:OnMouseWheel()` of `IKnobControlBase` and `ISliderControlBase` use `std::fmod` to round to the a stepped value. There are step values, for which this will not work.

Take e.g. a step value of 0.1: As we all know, 0.1 can not be presented exactly as a IEEE double value, which leads to the problem, that adding the step value 0.1 to a value of e.g. 0.4 will result again in a value of 0.4.

To understand the problem, look at the following small program:

```
#include <stdio.h>
#include <math.h>

int main()
{
    double x = 0.4;
    double s = 0.1;
    double x_s = x + s;
    double m = fmod(x_s, s);
    double new_x = x_s - m;
    printf("x = %.17g\ns = %.17g\nx+s = %.17g\nfmod(x+s,s) = %.17g\nx+s-fmod(x+s,s) = %.17g\n", 
            x, s, x_s, m, new_x);
}
```
The output is:

x = 0.40000000000000002
s = 0.10000000000000001
x+s = 0.5
fmod(x+s,s) = 0.099999999999999978
x+s-fmod(x+s,s) = 0.40000000000000002

The problem can be fixed by this small patch (using std::round instead of std::fmod):

```
diff --git a/IGraphics/IControl.cpp b/IGraphics/IControl.cpp
index 17fe1f78d..af5f816b1 100644
--- a/IGraphics/IControl.cpp
+++ b/IGraphics/IControl.cpp
@@ -842,7 +842,8 @@ void IKnobControlBase::OnMouseDrag(float x, float y, float dX, float dY, const I
       pParam->GetBounds(l, h);

       v = l + mMouseDragValue * range;
-      v = v - std::fmod(v, pParam->GetStep());
+      const double step = pParam->GetStep();
+      v = std::round(v / step) * step;
       v -= l;
       v /= range;
     }
@@ -871,7 +872,7 @@ void IKnobControlBase::OnMouseWheel(float x, float y, const IMouseMod& mod, floa
       const double step = pParam->GetStep();
       v += d > 0 ? step : -step;
       v = Clip(v, l, h);
-      v = v - std::fmod(v, step);
+      v = std::round(v / step) * step;
       v -= l;
       v /= range;
     }
@@ -976,7 +977,8 @@ void ISliderControlBase::OnMouseDrag(float x, float y, float dX, float dY, const
[...]

Issue #690, opened 2021-01-30
IKnob/SliderControlBase::OnMouseWheel rounding errors


gvampgh:
Due to rounding errors, Knobs & sliders get stuck at some circumstances.

**To Reproduce**
Use IPlugExamples on Windows
Change first parameter to GetParam(kParamGain)->InitInt("Gain", 100, 1, 100, "%");
Move sliders or knobs with **mouse wheel** under 28% and then above again, it stucks at 28%
The value is v = 28.999999999999996 at this line:

```
      v += d > 0 ? step : -step;
      v = Clip(v, l, h);
      **v = v - std::fmod(v, step);**
      v -= l;
      v /= range;
```

**Expected behaviour**
Normal mouse wheel behaviour

**Screenshots**
![grafik](https://user-images.githubusercontent.com/48643530/106365387-92aa8f80-6335-11eb-90b2-a26d609c42cc.png)

**IMPORTANT DETAILS**
* What plug-in format does it relate to - ALL
* What platform does it relate to - Windows (at least)
* What IGRAPHICS_BACKEND does it relate to, ALL

TBProAudio:
`v = v - std::fmod(v, step);` 
seems to be the problem.

```
#define __dround(x, n) (std::round((x) * pow(10.0, (n))) / pow(10.0, (n)))
v = __dround(v, log10(1.0 / step));  // step != 0
```
should do it better.

gvampgh:
This looks a little bit overdosed to me. :-) Adding a quarter of the step size here fixed it at least for my own UI and IPlugExamples (on Windows). I'm still not sure if this works for all combinations/step sizes and min/max values. BTW, I recognized that at least the ISliderControlBase has the same problem/needs the same fix.

```
      const double step = pParam->GetStep();
      **v += pParam->GetStep() * 0.25;**
      v += d > 0 ? step : -step;
```

PR #713, opened 2021-03-25
Repaired stepped controls

pthvogt:
The value is now rounded to the nearest stepped value, not more
to the next lower stepped value. The old behaviour did not work
always for fractional steps.

Resolves Issue #712

### DIFF
--- a/IGraphics/IControl.cpp
+++ b/IGraphics/IControl.cpp
@@ -842,7 +842,8 @@ void IKnobControlBase::OnMouseDrag(float x, float y, float dX, float dY, const I
       pParam->GetBounds(l, h);
 
       v = l + mMouseDragValue * range;
-      v = v - std::fmod(v, pParam->GetStep());
+      const double step = pParam->GetStep();
+      v = std::round(v / step) * step;
       v -= l;
       v /= range;
     }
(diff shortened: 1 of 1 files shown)


AlexHarker:
Would this also resolve #690?

AlexHarker:
I've just had a quick look at this and while I can see how the fix will likely work, the original code assumes that the parameter has a linear shape, which is not guaranteed. I think a more elegant and complete fix would simply be to use FromNormalised() and ToNormalized() on the parameter, which already includes this rounding, although then it would be constrained twice so I'll try to see if there's an easier way, or consider adding something to IParameter that correctly constrains normalised values.

I'll attempt to implement that and make a new proposal, but I think it'll be less code and will also operate correctly with parameters that don't have a linear shape.

AlexHarker:
Closing this due to #720 - thanks for the PR though - this was helpful in working through this, and the code was good - just noticed some other errors at the same time!

pthvogt:
Your PR is definitely better :-) Thank you!

AlexHarker:
No problem.

pthvogt:
Sorry, found just a bug in your commit :-(

Will create a new issue.

PR #715, opened 2021-03-25
Enabled self managed host bypass

### LINKED ISSUE #714: Self managed host bypass
iPlug2 manages the host bypass by itself, preventing `ProcessBlock()` to be called.

Sometime one wants to mange the bypass by oneself, e.g. to support a click free bypass or to update some GUI elements like a spectroscope.

Therefore we should enable the iPlug2 user to specify that `ProcessBlock()` should be called, regardless of the `GetPypassed()` state. If enabled `ProcessBlock()` has to call `GetBypassed()` and provide some code to bypass the processing.

PR for this will follow!

pthvogt:
iPlug2 manages the host bypass (VST3 and AUv2) by itself, preventing ProcessBlock() to be called.

Sometime one wants to mange the bypass by oneself, e.g. to support a click free bypass or to update some GUI elements like a spectroscope.

Therefore we should enable the iPlug2 user to specify that ProcessBlock() should be called, regardless of the GetPypassed() state. If enabled ProcessBlock() has to call GetBypassed() and provide some code to bypass the processing.

Resolves #714

### DIFF
--- a/IPlug/AUv2/IPlugAU.cpp
+++ b/IPlug/AUv2/IPlugAU.cpp
@@ -1732,7 +1732,7 @@ OSStatus IPlugAU::RenderProc(void* pPlug, AudioUnitRenderActionFlags* pFlags, co
       _this->SetChannelConnections(ERoute::kOutput, nConnected, totalNumChans - nConnected, false); // this will disconnect the channels that are on the unconnected buses
     }
 
-    if (_this->GetBypassed())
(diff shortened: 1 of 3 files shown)


olilarkin:
Thanks for the contribution. I feel that this is something that individuals can add if they need it. I'd rather not complicate the base classes so i'm going to chose not to make this design choice

PR #721, opened 2021-03-31
Allow paths to RPP files to contain spaces

jpauwels:
If iPlug2 is checked out in a directory whose path contains spaces, the Reaper projects don't open unless you quote the paths in the Xcode workspaces (see changes, same single line change applied to 48 files).


jpauwels:
Dear @olilarkin @AlexHarker could you have a look at this PR. It's as uncontroversial as can be and solves a real nuisance. I've just updated and rebased it on the latest main branch.

AlexHarker:
Thanks @jpauwels. It looks good to me and makes sense to change for duplicating projects - @olilarkin are there any issues here - is this likely to get overwritten if we pull in projects from the out of source repo at some point?

Issue #724, opened 2021-04-02
Tooltips are only enabled or disabled when the GUI is built


hor-net:
IGraphics::EnableTooltips has effect only if called when the GUI is build, if the GUI of the plugin is already open and you toggle the switch it doesn't really enable or disable the tooltips. If you close and open the GUI again the switch takes effect. 

**To Reproduce**
Simply create a button to enable or disable tooltips and toggle it, it has no effect until GUI is closed and re opened

**Expected behaviour**
Tooltips should be enabled or disabled as soon as the switch is toggled so the user can hide them or show them as a kind of online help

The issue applies to every plugin format, both on windows and mac (I have not tested linux) and for every IGRAPHICS backend

PR #725, opened 2021-04-02
Allow enabling or disabling tooltips when the GUI is open

hor-net:
fixes issue #724


olilarkin:
This doesn't look right to me in IGraphicsWin. The original code is also a bit nasty. I will investigate

AlexHarker:
We may also require a web implementation or alteration - not sure if tooltips are a thing there...

Issue #731, opened 2021-04-07
SetLatency() in VST3 fails in Ableton Live 10 - suggested fix


Fender19:
Issue discovered in Ableton Live with VST3 version: plugin fails on loading due to issue with SetLatency().

I checked iPlug1 SetLatency() and found it included a check for null ptr.  I added the check to IPlug2 IPlugVST3::SetLatency(int latency) and it fixed it.

IPlugVST3.cpp - CURRENTLY in iPLUG2:
```
void IPlugVST3::SetLatency(int latency)
{
  IPlugProcessor::SetLatency(latency);

  FUnknownPtr<IComponentHandler>handler(componentHandler);
  handler->restartComponent(kLatencyChanged);
}
```

IPlugVST3.cpp - SUGGESTED FIX:
```
void IPlugVST3::SetLatency(int latency)
{
  IPlugProcessor::SetLatency(latency);

  FUnknownPtr<IComponentHandler>handler(componentHandler);
  if (handler != nullptr)//add
  {                      //add
      handler->restartComponent(kLatencyChanged);
  }                      //add
}
```

TBProAudio:
There much more DAWs having problems with VST3 latency settings.
This why we
1) always `#define PLUG_LATENCY 0` in config.h and do the latency setting later
2) check componentHandler as well, as this could be also invalid:

```
void IPlugVST3::SetLatency(int latency)
{
  if (componentHandler != nullptr)                                // Check
  {
    FUnknownPtr<IComponentHandler>handler(componentHandler);
    if (handler != nullptr)                                               // Check
    {
      IPlugProcessor::SetLatency(latency);                       // Set IPlugBase here only, otherwise out of sync
      handler->restartComponent(kLatencyChanged);
    }
  }
}
```

Fender19:
TBPro - your code takes the suggested "fix" one step further (and better than what I proposed).  Currently there is no check for handler in the IPlugVST3.cpp function for SetLatency().

olilarkin:
merged

PR #732, opened 2021-04-08
Test for validity of componentHandler in IPlugVST3::SetLatency()

olilarkin:
from #731

### DIFF
--- a/IPlug/VST3/IPlugVST3.cpp
+++ b/IPlug/VST3/IPlugVST3.cpp
@@ -206,7 +206,7 @@ void IPlugVST3::EndInformHostOfParamChange(int idx)
 
 void IPlugVST3::InformHostOfParameterDetailsChange()
 {
-  FUnknownPtr<IComponentHandler>handler(componentHandler);
+  FUnknownPtr<IComponentHandler> handler(componentHandler);
   handler->restartComponent(kParamTitlesChanged);
 }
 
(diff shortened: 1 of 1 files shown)


AlexHarker:
If the test fails does the caller need to know, or alternatively, should we not call IPlugProcessor::SEtLatency in either case so that when the component restarts the latency is correct? Otherwise I can't see where the new latency would get registered.

TBProAudio:
You can always check in the plugin code (onparamchange, processblock, onidle) if (VST3) setlatency worked by calling getlatency(). If equal to "real" latency everything is fine :-)

TBProAudio:
... and VST2/VST3/AAX APIs (AU I dont know currently) return status (OK, NOK) of setlatency. So even this could be checked ...
Because acceptance of latency changes depends on DAW, plugin format and plugin/DAW state.

AlexHarker:
I'm trying to suggest whatever will minimise the need for the plug programmer to check through this - so a scenario that is simplest to use is best - to me having to call GetLatency() wouldn't achieve that, but we'll see what @olilarkin  says :-)

sstillwell:
I'll mention that we have an instance where failure to test for valid pointer is causing a crash when testing VST3 using the VST3 SDK 'validator' command - the plugin has variable latency that is impacted by a parameter change.

PR #735, opened 2021-04-26
Workaround for window resize loop bug per Justin Frankel @ Cockos, In…

### LINKED ISSUE #734: Window goes into resize loop when dragged across monitors in Windows
**CHECK**
Done

**Describe the bug**
When dragging a plugin window across monitors in REAPER, plugin window can go into a resize loop growing to enormous size

**To Reproduce**
Steps to reproduce the behaviour.
Configure a Windows system with multiple monitors, where at least one is HiDPI and another is standard DPI.  Drag plugin slowly across the border between those two monitors.  When the center point of the window crosses the boundary, a resize is triggered, which can move the center point back to the original monitor, which triggers a resize again...this can end up looping ad nauseam.

**Expected behaviour**
Plugin window should only resize once when dragged across monitor boundaries, regardless of that monitor's HiDPI status.

**Screenshots**

**IMPORTANT DETAILS**
* What plug-in format does it relate to - VST2, VST3, AUv2, AUv3, AAX, APP or ALL? - ALL
* What platform does it relate to - Windows/macOS/iOS/Linux or ALL? - Windows
* What IGRAPHICS_BACKEND does it relate to, IGRAPHICS_NANOVG, IGRAPHICS_SKIA or ALL? - Tested with SKIA, unknown for other backends.

**Additional context**

Will supply PR to work around per Justin Frankel of Cockos, Inc. (Author of REAPER)

sstillwell:
…c. - when resizing on multimonitor Windows, only resize when mouse released of window drag.

Fixes #734 

* [create an issue](https://github.com/iPlug2/iPlug2/issues/new/choose) first to outline the problem, to reference in this PR - Done
* clearly describe the problem that the PR fixes - works around multiple monitor resizing loop described in #734 by only resizing once when mouse is released
* each PR should address one thing. It can include multiple commits, but you should try and tidy up work done into logical units if possible. - Done

thanks, we like PRs!

oli & alex

### DIFF
--- a/IGraphics/Platforms/IGraphicsWin.cpp
+++ b/IGraphics/Platforms/IGraphicsWin.cpp
@@ -242,9 +242,12 @@ void IGraphicsWin::OnDisplayTimer(int vBlankCount)
   }
 
   // TODO: move this... listen to the right messages in windows for screen resolution changes, etc.
-  float scale = GetScaleForHWND(mPlugWnd);
-  if (scale != GetScreenScale())
-    SetScreenScale(scale);
+  if (!GetCapture()) // workaround Windows issues with window sizing during mouse move
+  {
+    float scale = GetScaleForHWND(mPlugWnd);
+    if (scale != GetScreenScale())
+      SetScreenScale(scale);
+  }
 
   // TODO: this is far too aggressive for slow drawing animations and data changing.  We need to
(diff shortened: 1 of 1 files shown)


olilarkin:
merged in bb88892c81c0f03d36086c6ef7b6e2af2a854a01

PR #752, opened 2021-06-02
Address -Wreorder warnings

crncnnr:
Hey,

Here's the first PR suppressing some of the compiler warnings enabled by `-Wall`.

I'm only compiling VST3 at the moment, so I can't guarantee I've covered all instances, but it's a start.

`-Wreorder` is a pretty minor complaint from the compiler: it prefers that the member initialiser lists match the initialisation order (feels more like a pedantic style issue if I'm being honest!).

I'm hoping at the end of this it'll also be possible to enable `-Wall` in the Xcode projects, if that's something considered desirable.

Thanks,
Ciarán :)

### DIFF
--- a/IGraphics/Controls/IBubbleControl.h
+++ b/IGraphics/Controls/IBubbleControl.h
@@ -50,9 +50,9 @@ public:
    * @param strokeColor The stroke color of the bubble */
   IBubbleControl(const IText& text = DEFAULT_TEXT.WithAlign(EAlign::Center), const IColor& fillColor = COLOR_WHITE, const IColor& strokeColor = COLOR_BLACK, float roundness = 5.f)
   : IControl(IRECT())
+  , mRoundness(roundness)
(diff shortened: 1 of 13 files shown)


AlexHarker:
Thanks for doing this not so fun stuff - looks pretty sane from a skim read so I imagine we'll want to merge in theory, although the windows builds are failing on CI at the moment which needs investigating.

crncnnr:
Hmm looks like it's failing with the same errors, but I'm unsure if it's related to these changes. Could it have something to do with the CI environment?

olilarkin:
> Hmm looks like it's failing with the same errors, but I'm unsure if it's related to these changes. Could it have something to do with the CI environment?

yep, something is wrong with the windows CI build. It builds fine locally

olilarkin:
msbuild bug in latest visual studio https://github.com/dotnet/msbuild/issues/6499

PR #758, opened 2021-06-12
set mMouseOutDuringDrag to false when drag-entering

### LINKED ISSUE #757: OnMouseOut being called although control was entered (Mac)
**Describe the bug**
A mouse up event triggers a `OnMouseOut` although the mouse is within the control area.

**To Reproduce**
1. drag a control
2. while dragging, leave the control area
3. still dragging, enter the control area again
4. release the mouse

**Expected behaviour**
`OnMouseOut` is not called upon `OnMouseUp`.

pressplay123:
Fixes #757 

The variable is already set to false within `mouseEntered`, but it should also (instead?) be set to false within `draggingEntered`.

### DIFF
--- a/IGraphics/Platforms/IGraphicsMac_view.mm
+++ b/IGraphics/Platforms/IGraphicsMac_view.mm
@@ -1218,6 +1218,8 @@ static void MakeCursorFromName(NSCursor*& cursor, const char *name)
 {
   NSPasteboard *pPasteBoard = [sender draggingPasteboard];
 
+  mMouseOutDuringDrag = false;
+
   if ([[pPasteBoard types] containsObject:NSFilenamesPboardType])
     return NSDragOperationGeneric;
   else


pressplay123:
Hm, this only partly fixes the issue. It only works if you stay within the window while dragging. I think that in addition, `draggingExited(_ sender: NSDraggingInfo?)` has to be defined (https://developer.apple.com/documentation/appkit/nsdraggingdestination/1416056-draggingexited). I will try adding that as well.

pressplay123:
No, `draggingExited` did not help. For some reason, `draggingEntered` is not called when you leave the window in between. I am a bit lost here. The original fix is still okay, but as I said, it does not solve the issue completely

AlexHarker:
draggingEntered is designed to respond to file drop (or content drop) interactions - it isn't going to help here I don't think:

https://developer.apple.com/documentation/appkit/nsdraggingdestination/1416019-draggingentered?language=objc

I'm confused as to why mouseEntered (where this flag is set - see the issue) isn't working for you

AlexHarker:
@pressplay123  AS this isn't a fix for the issue you were experiencing I'm going o close the PR but leave the issue open. If you can more fully explain the issue I'll look at this again, but as I've explained above I don't yet see why you're having an issue and draggingEntered is about file drops, not more generic mousing.

pressplay123:
@AlexHarker sorry, I totally forgot about this issue (which clearly indicates that it is not very crucial). I checked the behaviour again and still can observe the following:
- you press down the mouse on a control and start dragging
- you drag the cursor out of the plug-in window and - while still holding down the mouse button - drag it back on the control
- you release the mouse and `OnMouseOut` is called. Which is unexpected.

The call stack shows that the function was called because `mMouseOutDuringDrag` was still true (in `IGraphicsMac_view.mm`). Apparently, the flag has not been reset. This does not happen if the mouse remains inside the plug-in window while dragging.

If you have the time, could you please check if you can observe the same issue? Thanks

AlexHarker:
I will double check this bu as you can see on #757 I wasn't previously able to reproduce - can you give as specific a set of details of your test setup as possible including:

- Mac OS Version
- A iplug example that will show the problem (e.g. iPlugEffect)
- A plugin format
- A host

That way I can attempt to reproduce your setup as closely as possible and hopefully get to the bottom of why mMouseOutDuringDrag has not been set correctly (which should already be covered)

pressplay123:
The same behaviour can be reproduced on both Mojave and Big Sur. Also, it seems to be format-independent. To make it even more complicated, I cannot give you an example because the vector controls in iplug2 typically hide the cursor when being dragged which masks the issue. If you want to check anyway I would suggest to set a break point in a custom controls `OnMouseUp` and - following the description above - step out until you land in IGraphicsMac_view.

AlexHarker:
OK - I'm not sure why the behaviour wasn't reproducible on High Sierra, but I can reproduce on Big Sur.

The issue is that the mouse up fires before mouseEntered and hence the member variable is updated, but too late.

Looks like the fix is to add the NSTrackingEnabledDuringMouseDrag option to the tracking area - I'll aim to push that fix soon.

pressplay123:
Thank you very much. Yes, the description of this option sounds like the right fix for it.

PR #720, opened 2021-03-28
Graphics/parameter stepping

AlexHarker:
This PR adds a convenience method to IParam in order to allow normalised value to be constrained using stepping (for use in IControls).

It then also modifies OnMouseWheel() and OnMouseDrag() for sliders and knobs to correctly perform stepping using the parameter code - this reduces code, is simpler to read and fixes errors raised in two issues.

This should fix #690 and #712. It also replaces #713, which doesn't deal with parameter shapes correctly (the original code didn't either, so that is no fault of what was a useful PR to point at how to improve this).

I have tested this with drag and wheel for linear and non-linear parameters (knob and slider) and it appears to function correctly. A second look over it would be welcome.

PR #761, opened 2021-06-23
Repaired stepped controls (again :-) )

pthvogt:
Stepped parameter are buggy due to PR #720.

In IKnobControlBase::OnMouseWheel() and ISliderControlBase::OnMouseWheel() the difference d is applied twice (in different amounts).

Reason:

In both methods there exists an extra if-block for the stepped case. In the old code (before #720) the value of the parameter was fetched again at the begin of the if-block, in the new implementation we just use the already altered value.

In both if-blocks the line
v = pParam->FromNormalized(v);
is changed to
v = pParam->FromNormalized(GetValue());


pthvogt:
One comment: This is the version with minimal source code change. A better solution would be an `if() ... else` construct to avoid the double work in the stepped case. This would mean more testing and I have currently no time for this :-(

olilarkin:
Fixed in #07e3e44, thanks!

AlexHarker:
Thanks @olilarkin - that commit (from a skim read) looks good to me!

PR #764, opened 2021-06-26
Getting DAW transport info

John-Crosby:
Please add just two lines so we can access:
- Song position (number of beats) and 
- state of DAW transport (Play or Stop).

Thanks!


olilarkin:
added in a6aa44d32f2e1b87a68a75e0364864eae8ac859a

Issue #775, opened 2021-07-22
WebKit framework not found when building IPlugWebUI example project


remaininlight:
**Describe the bug**
When building the IPlugWebUI example project I get the following error: 'Framework not found Webkit'. 

This is caused by incorrect capitalization of Webkit in the XCode project, I have put a pull request in here fixing this:
https://github.com/iPlug2/iPlug2/pull/776

**To Reproduce**
Build the IPlugWebUI example project on a case sensitive file system.

**Expected behaviour**
The IPlugWebUI example project should build and run.

**Screenshots**
N/A

**IMPORTANT DETAILS**
* What plug-in format does it relate to - VST2, VST3, AUv2, AUv3, AAX, APP or ALL?
ALL
* What platform does it relate to - Windows/macOS/iOS/Linux or ALL?
Tested on macOS
* What IGRAPHICS_BACKEND does it relate to, IGRAPHICS_NANOVG, IGRAPHICS_SKIA or ALL?
N/A

**Additional context**

PR #776, opened 2021-07-22
Fixed capitalisation issue preventing linking for WebKit framework in…

remaininlight:
When building the IPlugWebUI example project I get the following error: 'Framework not found Webkit'. This is filed as an issue here:
https://github.com/iPlug2/iPlug2/issues/775

This is caused by incorrect capitalization of Webkit in the XCode project, which this pull request fixes.


olilarkin:
thanks, fixed in 6e273c3

PR #778, opened 2021-07-30
Rework AUv3 Support, add SwiftUI example

olilarkin:
This PR

- adds a huge amount of fixes for AUv3 on macOS
- reworks IPlugCocoaViewController behaviour for better swift-interop
- adds IPlugSwiftUI example, IPlugSwift becomes IPlugUIKit
- reworks IPlugPaths code to find resources better in sandbox and e.g. in Catalyst


olilarkin:
merged

Issue #800, opened 2021-12-09
IPlug2 AAX plugins do not report levels or gain reduction to the host


sstillwell:
AAX plugins support dedicated meters that may or may not be attached to controls in the plugin UI.  Sometimes these meters are intended to feed information back to the Pro Tools host UI or to a hardware control surface.

Pro Tools simply expects you to report the highest value found in any buffer for each block - the meter ballistics are handled by Pro Tools itself.

I propose a change to allow any instantiated AAX plugin to automatically have a single Input and Output meter, and for Dynamics plugins to have a Gain Reduction meter.  The primary use case for this is to report Gain Reduction in the Pro Tools mixer UI, where the gain reduction meter is just alongside the track's level meter.

I will submit a pull request with known working code - this was refactored from code in our legacy IPlug repository.

sstillwell:
I should mention that Input and Output meter values are automatically derived from code in IPlugAAX_Processor.cpp, but that the only thing that needs implementing in the host plugin is a single call to SetGRBufferMaxValue(gain_reduction_value) in the plugin's ProcessBlock() method with the highest gain reduction in that block (all channels/buses) in the range 0...1, with 0 being NO gain reduction, and 1 being maximum gain reduction (i.e. silence).

sstillwell:
After moving the code from IPlugProcessor to IPlugAAX, please note that the code to report gain reduction inside your plugin's ProcessBlock() method must be guarded to only compile on AAX, eg:

#ifdef AAX_API
SetGRBufferMaxValue(some_value);
#endif

PR #801, opened 2021-12-09
Stillwellaudio/aax meter pull request

sstillwell:
* Created issue #800 
* This pull requests adds input, output, and gain reduction meters (for Dynamics plugins) that report information back to Pro Tools for metering within the host UI - this could additionally be extended to other formats than AAX if the host supports it.

### DIFF
--- a/IPlug/IPlugProcessor.cpp
+++ b/IPlug/IPlugProcessor.cpp
@@ -483,10 +483,13 @@ void IPlugProcessor::AttachBuffers(ERoute direction, int idx, int n, PLUG_SAMPLE
 
 void IPlugProcessor::PassThroughBuffers(PLUG_SAMPLE_DST type, int nFrames)
 {
+  mMeterLevelIn = GetInputBufferMaxValue(nFrames);
+  mMeterLevelGR = 0.;
   if (mLatency && mLatencyDelay)
     mLatencyDelay->ProcessBlock(mScratchData[ERoute::kInput].Get(), mScratchData[ERoute::kOutput].Get(), nFrames);
   else
     IPlugProcessor::ProcessBlock(mScratchData[ERoute::kInput].Get(), mScratchData[ERoute::kOutput].Get(), nFrames);
+  mMeterLevelOut = GetOutputBufferMaxValue(nFrames);
 }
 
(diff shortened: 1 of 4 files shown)


hor-net:
I was waiting for something like this always wanted to work on it but never found the time! Thank you! one vote for me!

sstillwell:
@hor-net do be sure to read the notes on the linked issue #800 about how to implement Gain Reduction in Dynamics plugins - you'll need to add a single call in your plugin's ProcessBlock() method to report the highest gain reduction in that block (0.0 = no reduction, 1.0 = full reduction to silence).  That's per Pro Tools AAX SDK docs.

IPlug/IPlugProcessor.cpp
```diff
@@ -509,11 +512,16 @@ void IPlugProcessor::PassThroughBuffers(PLUG_SAMPLE_SRC type, int nFrames)
 
 void IPlugProcessor::ProcessBuffers(PLUG_SAMPLE_DST type, int nFrames)
 {
+  mMeterLevelIn = GetInputBufferMaxValue(nFrames);
```
olilarkin, line 515: This is going to affect all plug-in formats. This should move out of IPlugProcessor

sstillwell:
I'll do the work, but do you have a suggestion as to where?  I placed it there because it's able to surround the call to the plugin's ProcessBlock() method, and also to plan ahead if VST3 or other SDK would support similar mechanism in the future.

Again, willing to do the work, but unsure where else it could logically go.

olilarkin:
https://github.com/iPlug2/iPlug2/blob/master/IPlug/AAX/IPlugAAX.cpp#L366

sstillwell:
I'll work on it this evening if I have a chance.  Thanks!

sstillwell:
I've refactored into IPlugAAX.h/.cpp - it works here.  Please review @olilarkin.  It's not pretty but it works.

sstillwell:
FYI, making this change means that the call to SetGRBufferMaxValue() in the plugin's ProcessBlock() method must be guarded by:

#ifdef AAX_API
SetGRBufferMaxValue(some_value);
#endif

Issue #772, opened 2021-07-13
SetLatency() issue with VST3 in Ableton Live - merged #731 only partial fix


rexmartin158:
**Describe the bug**
Ableton Live does not set the correct latency per SetLatency() when a VST3 plugin is constructed.

**To Reproduce**

1. Create a blank Ableton Live project
2. Insert VST3 version of plugin on any track or bus - the plugin reports 0 samples latency regardless of SetLatency() value
3. Turn the plugin off ("deactivate") - now it reports the correct latency
4. Turn the plugin on again - still reports correct latency.
5. Save the project
6. Reload the project - the plugin loads up reporting 0 latency again. If you deactivate then re-activate the plugin it will report correct latency again, but every time you reload the project it will report 0 incorrectly again until you do this.

**Expected behaviour**
Plugin should load correct SetLatency() value on initial construction (SetLatency() called in OnReset()).

**Cause of the problem**
Fix for SetLatency() merged in #731 fixes the crash on startup but does not address the latency issue.  The reason is because Ableton does not have a "componentHandler" until after MakeGraphics() is totally finished.  Ableton calls OnReset() twice during construction but neither time is "componentHandler" a valid pointer so SetLatency() is never implemented.

**Workaround**
I don't know if this can be "fixed" in iPlug code because it appears to be a flaw in Ableton.  All other DAWs I have tested enact SetLatency() immediately on the first call to OnReset().  If it can be addressed via iPlug please advise.

In the meantime, as a workaround for Ableton I have added a latency check to OnActivate() as follows:

```
void MyPlugin::OnActivate(bool state)
{
    if (state)
    {
        if (GetHost() == kHostAbletonLive)
        {
            if (GetLatency() != desiredlatency) SetLatency(desiredlatency);
        }
****

```
Ableton calls OnActivate() one more time (a third time) at the conclusion of plugin construction at which point "componentHandler" is valid and this "hack" then loads the correct latency. 

Not sure if there are any side effects of this hack but it seems to fix the problem.

AlexHarker:
Thanks for this - if you remove your hack, but edit PlugVST3.cpp to place this line:

      IPlugProcessor::SetLatency(latency);

before the two if statements -does that also work? That looks like the correct fix.

AlexHarker:
I've tested and put the fix here:

https://github.com/iPlug2/iPlug2/tree/VST3/latency

I'll merge it soon but it'd be good to have it confirmed that this removes the need for a workaround.
[...]

PR #803, opened 2021-12-09
Correct iPlugVST3::SetLatency()

AlexHarker:
Addresses #772.

This moves the call to super such that it always occurs, regardless of whether the component handler exists. This means that calls to set the latency in the constructor of a plugin will not be ignored.


olilarkin (approved):
LGTM

PR #808, opened 2021-12-27
Fix memcpy not found error in some cases

luciascarlet:
This is a one-liner that fixes a problem I came across while playing around with trying to get iPlug to build via CMake. An error came up for me about "memcpy" not being declared, because it's a function in `<cstring>` which is never included. I was using a newer CXX standard (CXX20) so it might be related to this. I'm still new to CMake and C++ in general, so I'm not entirely sure what's going on here, but I felt this would be worth pointing out anyway.

### DIFF
--- a/WDL/heapbuf.h
+++ b/WDL/heapbuf.h
@@ -43,6 +43,7 @@
 #endif
 
 #include "wdltypes.h"
+#include <cstring>
 
 class WDL_HeapBuf
 {


olilarkin:
Hello, which iPlug2 file was the problem? heapbuf.h is part of WDL, which I try not to modify, better if we can include <cstring> before including heapbuf.h

FYI, there has been substantial CMake work done (on https://github.com/iPlug2/iPlug2/tree/linux)

luciascarlet:
For some reason this doesn't actually happen to me anymore after setting up a project completely from scratch again, weird. I'd imagine something got messed up for me when I tried to use the Linux branch the first time or something.

PR #809, opened 2021-12-28
Copy resources into .app bundle on macOS

luciascarlet:
Bundled application resources would previously be output into a separate folder called {PLUG_NAME}-app instead of the .app bundle, so the resulting .app bundle was not executable. I have changed the resource directory to {PLUG-NAME}.app/Resources/Contents, so that the app bundle is now executable and includes the icon and .nib required to run it.

Feel free to request edits if needed – I'm not sure whether changing the res_dir variable affects the Linux APP target or not. This *does* resolve the problem on macOS, though, so maybe a conditional should be used if it does affect Linux.

### DIFF
--- a/Scripts/cmake/APP.cmake
+++ b/Scripts/cmake/APP.cmake
@@ -120,7 +120,7 @@ iplug_source_tree(iPlug2_APP)
 macro(iplug_configure_app target)
   iplug_target_add(${target} PUBLIC LINK iPlug2_APP)
 
-  set(res_dir "${CMAKE_BINARY_DIR}/${PLUG_NAME}-app/resources")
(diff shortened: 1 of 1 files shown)


olilarkin:
@Bindernews are you still with us and would you like to have a look at this ? :-)

Bindernews:
Yeah, I'm still alive. Took a break from coding for a few months. Finally starting to get back into it, so good timing on this. It's fine to merge. I need to re-open that project and see what's what, because I think I have several changes that should make life easier for the CMake users. I guess we'll find out.

kmturley:
Every other plugin framework uses cmake for builds. I think it's essential work you're doing!

Check out the CI commands for each framework i've been compiling here:
https://github.com/studiorack/studiorack-workflows/tree/main/.github/workflows

Issue #816, opened 2022-01-11
IPlugQueue ElementSize function does not work correctly


Youlean:
IPlugQueue ElementSize function does not work correctly.

```
  size_t ElementsAvailable() const
  {    
    return (mWriteIndex.load(std::memory_order_acquire) - mReadIndex.load(std::memory_order_relaxed))%mData.GetSize();
  }
```

For example consider:
```
          size_t size = 5;
          
          size_t write = 0;
          size_t read = 4;
```

if the read index is higher than write index, size will be wrong because both read and write is size_t thus it will produce a bad result i.e.

`size_t(0) - size_t(5) = some huge number`

This should be the fix:
```
  size_t ElementsAvailable() const
  {
    size_t write = mWriteIndex.load(std::memory_order_acquire);
    size_t read = mReadIndex.load(std::memory_order_relaxed);

    if (read > write)
    {
      return mData.GetSize() - (read - write);
    }

    return write - read;
  }
```

AlexHarker:
Thanks. I've tested the original and the proposed fix (slightly refactored for brevity) using the below code.

The original fails, but the fix works (this is a single threaded test, but the mechanism for accessing the values is the same, so this should exhibit the same threading behaviour as before).

```
#include "IPlugQueue.h"

#include <iostream>
#include <random>

std::random_device rd;
std::mt19937 gen(rd());
    
int randInt(int lo, int hi)
{
    std::uniform_int_distribution<> distr(lo, hi);
    return distr(gen);
}

bool pushItems(iplug::IPlugQueue<int>& queue, int size, int& inQueue)
{
    int add = randInt(0, size - inQueue);
    
    for (int i = 0; i < add; i++, inQueue++)
        queue.Push(randInt(0, 4096));
    
    return queue.ElementsAvailable() != inQueue;
}

bool popItems(iplug::IPlugQueue<int>& queue, int size, int& inQueue)
{
    int pop = randInt(0, inQueue);
    int blank;
    
    for (int i = 0; i < pop; i++, inQueue--)
        queue.Pop(blank);
    
    return queue.ElementsAvailable() != inQueue;
}

bool testQueue(int testSize)
{
    int size = randInt(12, 512);
    int inQueue = 0;
    iplug::IPlugQueue<int> queue(size);
    
    for (int i = 0; i < testSize; i++)
    {
        if (pushItems(queue, size, inQueue))
            return false;
        if (popItems(queue, size, inQueue))
            return false;
    }
    
[...]

PR #817, opened 2022-01-12
Fix IPlugQueue<>::ElementsAvailable()

AlexHarker:
This addresses #816 as proposed in the issue, but with a small refactor.

Tested with the code posted in the issue, but I' appreciate @olilarkin taking a look.


olilarkin:
lgtm

Issue #818, opened 2022-01-12
OnMouseDblClick isnt called on double click with right mouse button (skia, standalone, winx64)


janhase88:
OnMouseDblClick isnt called on double click with right mouse button (skia, standalone, win10x64)

possible fix:
IGraphicsWin.cpp on Line 466 add:  `case WM_RBUTTONDBLCLK:`

AlexHarker:
This is also the case for the Mac platform according to the code (but not the web platform) - we will need to decide what the correct functionality is.

olilarkin:
@olilarkin will add for mac and win

PR #820, opened 2022-01-14
detection of double click for right mouse button (Win)

janhase88:
https://github.com/iPlug2/iPlug2/issues/818


olilarkin:
Thanks for the PR. We will make a new PR adding this for mac as well as windows

PR #826, opened 2022-01-24
Linux bl fix

nd-bl:

thanks, we like PRs!

oli & alex


nd-bl:
Hello, 
I previously developed the BlueLab plugins using iPlug (the new betas were developed using iPlug2).
The branch is not finished but there is some long work in it, pushing it as it is, hope this can help.
Nicolas

olilarkin:
Thanks @nd-bl , I changed the PR to merge to the linux branch. Your bluelab plug-ins look very interesting. I'm curious to know what you're doing with them in the future, msg me via contact AT olilarkin.co.uk if you like

PR #828, opened 2022-01-28
Use CatmullRom() interpolation

AlexHarker:
This cubic interpolator is far faster than the one we were using before on skia cpu and still looks good to me.


pthvogt:
Funny, are you sure? These two options should use the same algorithm, only with different parameters. (Just looked in Skia header file…) CatmullRom has a little more ringing artifacts, but is a little sharper.

Having some history of using image interpolation in research, I never could decide, which is better :-)

AlexHarker:
There is absolutely no doubt on the Mac that the speed difference is very significant.

 I note that some of the parameters in CatmullRom() are zero and from what I could see in the headers it is not possible to be sure that they take the same branches in the actual code, although yes, it does look like just a change of parameters. My assumption is that the zero parameters allow a quicker branch to be taken in practice. 

I can provide a test plugin for Mac if anyone wants to confirm, but it is the difference between a very slow sluggish UX and a snappy responsive UI for me.

PR #829, opened 2022-01-28
Au tidy

AlexHarker:
A couple of things tidied, a macro fix and a code reduction


olilarkin:
oops i squashed it. oh well

PR #834, opened 2022-02-12
Set interpolation more suitably for skia CPU and other versions of skia

AlexHarker:
On some testing it seems that CatMullRom() was only fast for one speed. It is much slower on CPU than Mitchell() for other speeds. On Metal etc. the speed seems fine in any case, so I propose using bilinear resampling on CPU (which is fast) and Mitchell() on other versions of Skia.


AlexHarker:
@pthvogt - you might be interested to follow this, given previous comments on the last PR.

olilarkin:
I tried it with the IGraphicsStressTest and confirmed

Issue #841, opened 2022-02-25
IGraphics::mInTextEntry invalid when platform text entry is dismissed


olilarkin:
If a platform text entry is dismissed, e.g. by moving the mouse wheel on a control when a text entry is open, IGraphics::mInTextEntry is left pointing to the control that triggered the text entry

olilarkin:
https://github.com/iPlug2/iPlug2/tree/graphics/text-entry-dismissal

PR #842, opened 2022-02-25
Clear IGraphics::mInTextEntry when platform text entry is dismissed

olilarkin:
This PR fixes https://github.com/iPlug2/iPlug2/issues/841

### DIFF
--- a/IGraphics/Controls/ITextEntryControl.cpp
+++ b/IGraphics/Controls/ITextEntryControl.cpp
@@ -544,7 +544,7 @@ void ITextEntryControl::DismissEdit()
 {
   mEditing = false;
   SetTargetAndDrawRECTs(IRECT());
-  GetUI()->ClearTextEntryControl();
+  GetUI()->ClearInTextEntryControl();
   GetUI()->SetAllControlsDirty();
 }
 
--- a/IGraphics/Controls/ITextEntryControl.cpp
+++ b/IGraphics/Controls/ITextEntryControl.cpp
@@ -544,7 +544,7 @@ void ITextEntryControl::DismissEdit()
 {
   mEditing = false;
   SetTargetAndDrawRECTs(IRECT());
-  GetUI()->mInTextEntry = nullptr;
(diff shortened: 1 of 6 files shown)


AlexHarker:
Thanks @olilarkin.

This fixes some significant parts of the issue, but in my case (in plugs where I'm hiding text display when the edit is in progress) it doesn't fix everything, because I store a bool when editing starts that I don't clear until SetValueFromUserInput() is called on my control.

There are a few options here:
- leave this as it is and I instead call GetUI->GetControlInTextEntry() to see if the control in question is in an edit (fixes my issues, but risks other iplug2 devs relying on a call to SetValueFromUserInput() that never comes.
- Always call SetValueFromUserInput() even if dismissing
- Add a new method to an IControl for when the entry is dismissed.

I'm not sure how common the pattern I've used is likely to be, but whilst we are looking at this we might as well consider the possible other scenarios.

AlexHarker:
We could also (as a separate decision) add a GetInTextEntry() const() to the IControl class.

olilarkin:
Since none of the example code exhibits this issue, I think you should go for option 1 for now.

AlexHarker:
OK - would you object to IControl helpers to check if they are in a text entry (and possibly also a pop-up, although I don't need that and I'm not sure if we can track that)

olilarkin:
Can you show me some code to reproduce the issue either in SRM repo or with a small example?

AlexHarker:
Well, I'm not sure full code is needed but DraggableBase is where the issue is.

I have mInEdit which is set true when I start the TextEntry and false in SetValueFromUserInput(). I've already adapted to not have a local and instead I've added to a base class this:

 ```bool GetInTextEntry() const { return GetUI()->GetControlInTextEntry() == this; }```

Something like that in IControl might be handy.

AlexHarker:
So the issue here is that I assume that if I start a text entry then I'll get a matching call to SetValueFromUserInput(), but that assumption isn't correct. I'm wondering if there are other cases where people need to know if the control is in a text entry.

PR #854, opened 2022-03-13
Feature ondrop xy

kcbanner:
In some cases (like dropping onto a IVTabSwitchControl subclass), you want to know the screen coordinates of the drop so you can determine the underlying button. When dragging into a control when the window doesn't have focus, the first event we get is the drop, and not a mouse move - so we can't track the button for the drop ahead of time.

This change just wires up the x,y coordinate into OnDrop. Instead of modifying the existing OnDrop, I added a new version that calls the old one as default impl, as to not break existing code, but if it's preferable just to modify it, I can change this.

### DIFF
--- a/IGraphics/Controls/Test/TestDropShadowControl.h
+++ b/IGraphics/Controls/Test/TestDropShadowControl.h
@@ -48,7 +48,7 @@ public:
     g.DrawLayer(mLayer);
   }
 
-  void OnDrop(const char* str) override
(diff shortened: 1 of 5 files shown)


olilarkin:
Thanks for the PR. It seems that drag and drop can be improved by responding to more OS calls. I am doing that work here:

https://github.com/iPlug2/iPlug2/tree/graphics/improve-drag-drop

i will close this PR, but thanks for the nudge!

PR #855, opened 2022-03-16
fix Prepare_resources-mac.py for python3 for new macOs 12.3 update

wcrhlo:
converted py3

Problem on MacOs 12.3 https://iplug2.discourse.group/t/macos-12-3-build/592

Also

Screenshot 2022-03-16 at 19 36 22

I'm not good at helping, but these fixes solve the build problem on the new mac Os 12.3 update. The bottom line is that after the update, the build stopped happening. If this is important, then you can translate all the files to Python3


olilarkin:
Thanks. There are quite a few more things to fix, which i have done in #857

PR #873, opened 2022-05-14
Linux 2022 fixes

StephanWidor:
- readd `#endif`that somehow got lost in merge of master branch
- add stddef.h include to WDL so that gcc knows `NULL` and `size_t`
-  fix resources path for standalone app

### DIFF
--- a/Scripts/cmake/APP.cmake
+++ b/Scripts/cmake/APP.cmake
@@ -120,9 +120,8 @@ iplug_source_tree(iPlug2_APP)
 macro(iplug_configure_app target)
   iplug_target_add(${target} PUBLIC LINK iPlug2_APP)
 
-  set(res_dir "${CMAKE_BINARY_DIR}/${PLUG_NAME}.app/Contents/Resources")
-
   if (WIN32)
+    set(res_dir "${CMAKE_BINARY_DIR}/${PLUG_NAME}.app/Contents/Resources")
     set_target_properties(${target} PROPERTIES
       OUTPUT_NAME "${PLUG_NAME}"
       RUNTIME_OUTPUT_DIRECTORY "${PLUG_NAME}-app"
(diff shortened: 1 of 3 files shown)


olilarkin:
Thanks for this. I don't like to modify the WDL files if possible. I wonder if that include can be fixed elsewhere, or submitted as a fix to [WDL](https://github.com/justinfrankel/WDL) which we then roll into iPlug2

StephanWidor:
> Thanks for this. I don't like to modify the WDL files if possible. I wonder if that include can be fixed elsewhere, or submitted as a fix to [WDL](https://github.com/justinfrankel/WDL) which we then roll into iPlug2

Good point, wasn't aware of that! I made a PR for WDL. How would that go in here then? Just copy&paste?
As a less intrusive alternative, stddef could also be included in IGraphicsLinux.cpp, just before including wdlutf8.h. What do you think?

olilarkin:
I think it should be fixed in IGraphicsLinux.cpp, IIRC i had a conversation with justin a long time ago about a similar thing

StephanWidor:
> I think it should be fixed in IGraphicsLinux.cpp, IIRC i had a conversation with justin a long time ago about a similar thing

ok, removed the include from wdltypes.h, and added it to IGraphicsLinux.cpp.

Scripts/cmake/APP.cmake
```diff
@@ -120,9 +120,8 @@ iplug_source_tree(iPlug2_APP)
 macro(iplug_configure_app target)
   iplug_target_add(${target} PUBLIC LINK iPlug2_APP)
 
-  set(res_dir "${CMAKE_BINARY_DIR}/${PLUG_NAME}.app/Contents/Resources")
-
   if (WIN32)
+    set(res_dir "${CMAKE_BINARY_DIR}/${PLUG_NAME}.app/Contents/Resources")
```
olilarkin, line 124: This looks weird for WIN32

StephanWidor: yep, was also a bit astonished, but this is what was set for win also before. I didn't want to change anything here, as I don't have a windows machine to test.

olilarkin: i removed it for now

PR #878, opened 2022-06-30
enhance CLAP support

Full-Bucket:
Enhance CLAP support:
* no crash when loading bitmaps
* report audio ports correctly
* react on (note) events

### DIFF
--- a/IPlug/CLAP/IPlugCLAP.cpp
+++ b/IPlug/CLAP/IPlugCLAP.cpp
@@ -186,11 +186,13 @@ clap_process_status IPlugCLAP::process(const clap_process *process) noexcept
   int nIns = process->audio_inputs->channel_count;
   int nOuts = process->audio_outputs->channel_count;
   int nFrames = process->frames_count;
-  
-  bool format64 = process->audio_inputs->data64;
-  
+
+  // does not seem to be safe...
+  //bool format64 = process->audio_inputs->data64;
+  bool format64 = process->audio_outputs->data64;
+
   assert(format64 == (bool) process->audio_outputs->data64);
-  
+
   SetChannelConnections(ERoute::kInput, 0, MaxNChannels(ERoute::kInput), false);
   SetChannelConnections(ERoute::kInput, 0, nIns, true);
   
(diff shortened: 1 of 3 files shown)


olilarkin:
Thanks for this. I've reformatted your PR into individual commits and merged to the CLAP branch. Also, be careful not to change the whitespace when you do a PR, since it makes it much more work for us to merge.

https://github.com/iPlug2/iPlug2/commits/clap

Thanks again!

Full-Bucket:
I have to thank you, Oli!

IPlug/CLAP/IPlugCLAP.cpp
```diff
@@ -441,18 +474,23 @@ void IPlugCLAP::ProcessInputEvents(const clap_input_events *in_events) noexcept
...
         {
           auto paramValue = ClapEventCast<clap_event_param_value>(event);
           
           int paramIdx = paramValue->param_id;
-          double value = paramValue->value;
-          
-          GetParam(paramIdx)->Set(value);
-          SendParameterValueFromAPI(paramIdx, value, false);
-          OnParamChange(paramIdx, EParamSource::kHost, event->time);
+          if (paramIdx >= 0 && paramIdx < this->NParams())
```
AlexHarker, line 485: Why is this check needed?

AlexHarker:
I think there was something else in here but I don't understand why it would be needed anyway... I've added a comment in case it is something we should have.

AlexHarker:
Thanks for looking over this - as @olilarkin mentioned it's in progress but the checking is welcome.

IPlug/CLAP/IPlugCLAP.cpp
Full-Bucket: I got some crazy (obviously uninitialized) values from Bitwig 4.3 so I thought it might be a good idea... ;-)

AlexHarker: I'm unclear on why the paramIdx should be out of range - it's also possible that we should be reporting this as host misbehaviour if it does happen, but I've not seen it to date - I would expect it to crash the plugin if it did happen.

AlexHarker: From what I recall the params are reported and the ids are arbitrary, so the host should just pass back values that have been passed to it from the plugin.

PR #888, opened 2022-08-11
Fix potential crash in IPlugCLAP::process()

Full-Bucket:
Leverages a crash in Bitwig when loading a CLAP plug-in


AlexHarker:
Thanks - I'm looking at audio ports at the moment.

@Full-Bucket - Can you explain the crash a little better - I *think* I know what might be happening, but I can't be sure.

It looks like there could be a situation in which format64 gets set on an empty audio bus and therefore ends up incorrect - is that right? Most likely that would be a plug with no inputs which incorrectly identifies the inputs as 32 bit because process->audio_inputs->data64 is a NULL pointer.

I may fix this slightly differently, and this code is likely to change for other reasons, so it's important to understand exactly how the crash comes about, rather than simply having a fix.

FWIW - in future IPlugCLAP will require all buses to be either 32 bit or 64 bit (which it hasn't up until now), which will firm up another possible source of audio connection issues.

Full-Bucket:
> It looks like there could be a situation in which format64 gets set on an empty audio bus and therefore ends up incorrect - is that right? Most likely that would be a plug with no inputs which incorrectly identifies the inputs as 32 bit because process->audio_inputs->data64 is a NULL pointer.
Yes, it crashed for an instrument plug-in without input ports. The problem ist that one must not assume process->audio_inputs to be NULL in this case.

AlexHarker:
Thanks.

@Full-Bucket - Can you please test a7b4b4e2da4f1ddc6c5439a7ac5d458e6695c76d

The issue should be fixed here, but you will require your version of clap-helpers to be taken from here: https://github.com/iPlug2/clap-helpers

Assuming that this commit works for you I will close this PR. The commit contains the changes that will ultimately support side chains and complex channel setups. There are still some things to test as it doesn't look like any available host actually checks configs at the moment (and I'm a little uncertain about the best thing to do as default when a config hasn't been selected) but it's necessary work to get where we need to be eventually.

AlexHarker:
As an update you can now use the official version of clap-helpers again, but you need to be at the tip as the changes we require are now merged. Thanks

AlexHarker:
Excellent - glad to hear it.

Issue #890, opened 2022-08-26
Remove the special entry in the index for CLAP submodules on the cmake-branch


grahnfan:
**Describe the bug**
When cloning the repo and updating submodules on the cmake branch
An error occurs with fetching submodules of CLAP that someone probably has stored in their local git index and 
accidently commit.

From https://github.com/iPlug2/iPlug2
   * branch            ba157df2f7ba83b13ccd0fd62dde711ba8e6add7 -> FETCH_HEAD
  Submodule path 'iPlug2': checked out 'ba157df2f7ba83b13ccd0fd62dde711ba8e6add7'
  Error: fatal: No url found for submodule path 'iPlug2/Dependencies/IPlug/CLAP_HELPERS' in .gitmodules
  Error: fatal: Failed to recurse into submodule path 'iPlug2'
  Error: The process '/usr/local/bin/git' failed with exit code 128

**To Reproduce**
1. clone iPlug2 repo
2. run git submodule update --init --recursive 

**Expected behaviour**
The CLAP_HELPERS and CLAP_SDK should not be stored in the modules/index unless they are specified as submodules
in a .gitmodules file and provided with a proper url

**Screenshots**

**IMPORTANT DETAILS**
Relates to ALL platforms
**Additional context**

AlexHarker:
The CLAP_HELPERS and CLAP_SDK were included in the clap branch which is for work in progress on clap support. It was set up for me to work on and intended to be carefully merged into master once the work is complete (which may not involve submodules, as other bits of IPlug2 are not done that way, but require a different approach, such as manual cloning.

I'm not sure how the cmake branch has come into being, but it looks like it has merged in changes for the clap branch (including very recent changes), which may or may not be in a finally mergeable state.

@olilarkin - can you shed any light on what's happened in terms of the history so we can ensure the correct solution?

olilarkin:
cmake and clap brances are both WIP and will be merged carefully. however thanks for the heads up

PR #891, opened 2022-08-26
Remove the special entry in the index for CLAP submodules on the cmak…

grahnfan:
This PR refers to #890 

The CLAP_HELPERS and CLAP_SDK should not be stored in the modules/index unless they are specified as submodules
in a .gitmodules file and provided with a proper url


olilarkin:
Thanks, i did merge. This Cmake branch is WIP - i want to merge at the same time as CLAP support

PR #892, opened 2022-08-28
Clap shaped parameters

AlexHarker:
This PR concerns the correct implementation of shaped parameters (via automation and host GUI) for CLAP.

Unlike the previous implementation this version reports all double parameters as normalised (0-1) and then applies the correct transformations for shaping to occur. This will break any saved sessions or automations but is probably the neatest implantation that works (bools/enum/ints must be reported in their correct ranges, unlike for VST3 because otherwise they cannot be stepped as far as the host is concerned). Branching for only non-linearly shaped parameters will be more expensive if it is even possible) as it requires a dynamic_cast and so making the choice based on parameter type seems the cleanest way.

I'll review this with @olilarkin.

However, it's possible @Full-Bucket and/or @heebje might want to check this out and be aware of how this will break previous versions. Sorry about that, but this is pretty likely to get merged, as I think having correct parameter shaping from the host is important.

### DIFF
--- a/IPlug/CLAP/IPlugCLAP.cpp
+++ b/IPlug/CLAP/IPlugCLAP.cpp
@@ -488,8 +488,8 @@ void IPlugCLAP::ProcessInputEvents(const clap_input_events *inputEvents) noexcep
           int paramIdx = paramValue->param_id;
           double value = paramValue->value;
           
-          GetParam(paramIdx)->Set(value);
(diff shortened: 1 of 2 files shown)


AlexHarker:
As there have been no comments in the last two weeks I'm going to merge this.

@Full-Bucket / @heebje - as you've been active with the CLAP stuff please be aware that this is a breaking change.

PR #893, opened 2022-09-04
Add Async File Dialogs & Popup Menus on macOS & Implement PromptForFile on iOS 

olilarkin:
This PR means that open/save panels and popup menus can be launched without freezing the plug-in UI on macOS *when the dialog is triggered inside e.g. IControl::OnEndAnimation(). It also adds the iOS implementation and adapts the other platforms


olilarkin:
@AlexHarker if you have time, would you be able to give this a once-over, since it modifies some code that has been around for a while?

olilarkin:
@AlexHarker I tried this with SRM plugs and didn't spot any problems

PR #896, opened 2022-09-14
Add child controls

olilarkin:
This PR adds IContainerBase - an IControl sub class that can be used to implement controls that nest sub controls. It adapts some IControls that previously managed sub controls in an unpleasant way


olilarkin:
splitting into smaller PRs, so closing

PR #898, opened 2022-09-20
Prevent crash on re-calling clap_deinit

Full-Bucket:
Prevent crash when trying to delete gPluginDesc on repeated calls of clap_deinit.
This happened with the latest v7.5 of host "Metro" by Sagan Technologies.


AlexHarker:
Thanks for pointing out this issue. I'm closing, because I decided it would be better to fix this another way but it should now be fixed.

The fix is here d1dcfed9c9423cbec869d03265591f540fe76fab

Issue #899, opened 2022-09-22
Lower case accelerate include causes build failure on Mac OS


andybest:
The recent include of the accelerate framework in `ISender.h` seems to cause problems when building for me due to the lowercase `#include <accelerate/accelerate.h>`

andybest:
Fixed by https://github.com/iPlug2/iPlug2/pull/900

PR #900, opened 2022-09-22
Capitalize Accelerate include

andybest:
This PR fixes https://github.com/iPlug2/iPlug2/issues/899 by changing the Accelerate include to be capitalized


AlexHarker (commented):
This is more correct anyway.

AlexHarker (approved):

Issue #903, opened 2022-10-25
Handle OpenGL failure on windows without crashing


hor-net:
When a windows machine creates the GUI for an IPlug2 plugin (or app) that requires OpenGL and the windows driver is not correctly configured (the default microsoft driver only provides OpenGL 1.1) it crashes badly. IPlug2 should handle this case informing the user that a valid OpenGL driver is required

**Step to reproduce**
Build any example using nanovg backend with OpenGL2, then chose the standard windows graphic driver and see the crash happen

**Expected behaviour**
The plugin should not crash but inform the user that a valid OpenGL driver is required

PR #904, opened 2022-10-25
handles failure for not having the correct OpenGL version in windows

hor-net:
fixes issue #903


olilarkin:
Thanks, working on this #1131

PR #912, opened 2022-11-27
AUv2 - Tail time property should not be writable

### LINKED ISSUE #911: AUv2s with tails fail validation on Ventura
If an iPlug2 plugin has a tail time then under Mac OPS Ventura it fails validation.

The test that fails concerns the fact that the tail time property returns as writable (which it should not)

AlexHarker:
This PR fixes #911


olilarkin (commented):
LGTM!

Issue #913, opened 2022-11-29
PromptForFile window kept behind plugin window on mac


svantana:
On mac, the NSPanel created by PromptForFile can be stuck behind the plugin window. Apple's documentation states vaguely that this behaviour can be dependent on the window size and other factors.

Tested on AU and VST plugin GUIs, Ableton Live 10.6, MacOS 12.6, 2019 Intel Macbook.

The problem seems to go away when adding `setFloatingPanel: YES`. (see accompaning PR).

olilarkin:
Thanks, I think I introduced this bug in https://github.com/iPlug2/iPlug2/commit/b71e9c29cfade09db3071332c7132d96ea57c104, and that it's not an issue if you supply the IFileDialogCompletionHandlerFunc for async operation. i'll investigate soon

olilarkin:
danke

PR #914, opened 2022-11-29
Force File NSPanels to be floating

svantana:
Solves issue #913


olilarkin (approved):
Thanks, this does indeed fix it. I'll also add to PromptForDirectory()

Issue #919, opened 2023-01-03
DEFAULT_FONT would be more convenient to set in one location.


daniel-p-gonzalez:
Currently it needs to be set per platform in the project configs. It would be nice to be able to set it in one location, e.g. in config.h.

olilarkin:
As mentioned on discord, the way to customize DEFAULT_FONT, if necessary, is to specify it as a preprocessor macro. You can do that for all targets by adding it to ALL_DEFS, e.g. add DEFAULT_FONT="Arial" to the end of this line

https://github.com/iPlug2/iPlug2/blob/master/Examples/IPlugEffect/config/IPlugEffect-mac.xcconfig#L27

daniel-p-gonzalez:
That's for Mac. You also need to add it for the other platforms, so you have to add it redundantly. I would also argue that it's much more of a config data thing, like app dimensions and vst3 categories (which live in config.h) for example. It's not at all like setting the GUI backend (which was mentioned on Discord), where it makes sense to put it in the project configs since it affects the build, rather than being config data.

I respect the difference of opinion though, and will just keep it that way in my fork. :)

PR #920, opened 2023-01-03
Include config.h in IGraphicsConstants.h to allow setting DEFAULT_FONT

daniel-p-gonzalez:
This allows setting DEFAULT_FONT in one location for convenience.

Fixes https://github.com/iPlug2/iPlug2/issues/919


olilarkin:
Thanks for contributing, but see my comment on the PR for the correct way to do this. We don't want the stuff in config.h to be exposed to all the IGraphics source code, that is not how it is designed

PR #929, opened 2023-02-21
Update zlib dependency to zlib 1.2.13

krig:
The 1.2.11 download is no longer available from zlib.net, so in order to make the download script work the simplest solution is to update to the 1.2.13 release.


olilarkin (approved):
LGTM

PR #930, opened 2023-02-21
Fix IGraphics build of freetype on Mac OS Monterey

krig:
When building freetype, the configure script was unable to find the stdlib headers (stdio.h, string.h) and the build fails.

To resolve this, it is necessary to set the CPATH env variable to the correct location for the standard headers.


olilarkin:
Thanks for the PR. I can compile a universal libfreetype.a on macOS Ventura on an intel machine but it fails on if I try and build it on apple silicon (#933). I tried this change but it didn't seem to make a difference. will leave this open for now

krig:
I am on apple silicon as well. There are other problems that prevent it from building on apple silicon, but this was one of the fixes that I did that was simple enough to PR. I had do so some major surgery to remove some x86 cross-compilation to get things to build entirely, unfortunately.

PR #931, opened 2023-02-24
cmake generated files for the workflows in in iPlugOOS template 

erikgrahn13:
cmake generated files for the workflows in iPlugOOS template to run.

needed to add both pdb and dsym generation in cmake for it to be able to use the makedist-win and makedist-mac scripts during worklows for iPlugOOS. 

Some formatting of the code was also made, that's why the diff looks bigger than it is


olilarkin:
Thanks for the PR! Is there any chance you can redo the PR without the whitespace changes, it's hard to see what changed

erikgrahn13:
of course! I can squash all commits together as well

PR #932, opened 2023-02-25
cmake generated files for the workflows in iPlugOOS template porject

erikgrahn13:
cmake generated files for the workflows in iPlugOOS template to run.

needed to add both pdb and dsym generation in cmake for it to be able to use the makedist-win and makedist-mac scripts during worklows for iPlugOOS.

Some formatting of the code was also made, that's why the diff looks bigger than it is


olilarkin:
thanks, merging these changes in 00e50987bdfb96b220629d19659f49cd7b125f4d

PR #934, opened 2023-02-27
Link ICoreWebView2 statically

olilarkin:
This PR changes the example projects using the MS ICoreWebView2 to link statically. It adapts the IWebView and IWebViewControl interfaces so that they are not passed paths to the dll or tmp folder.


hor-net:
Oh well it seems some of the PR I was preparing is useless now :D

Issue #700, opened 2021-02-27
Update to latest HIIR incl. SIMD


olilarkin:
HIIR in iPlug2 is old and doesn't include or use SIMD optimizations

Oversampler.h should be adapted to work with that for intel and arm

http://ldesoras.free.fr/src/hiir-1.33.zip

olilarkin:
or possibly https://github.com/unevens/hiir

PR #943, opened 2023-03-26
Apply Clang Format

olilarkin:
This PR applies clang-format to the IPlug2 codebase


AlexHarker:
Do we really want:

`ConstructorInitializerIndentWidth: 2`

rather than setting this to zero (which would match the previous style?). I can live with it, but it's not my preference

AlexHarker:
Do we want to format 3rd party things like iPlug/Extras/HIIR/*?

IGraphics/IGraphicsStructs.h
```diff
@@ -2526,45 +2560,120 @@ struct IVStyle
...
+    : colorSpec(colors)
+  {
+  }
+
+  IVStyle WithShowLabel(bool show = true) const
+  {
+    IVStyle newStyle = *this;
+    newStyle.showLabel = show;
+    return newStyle;
+  }
```
AlexHarker, line 2592: This is an issue, because it hasn't added a blank line when expanding these helper functions onto different lines

olilarkin: yeah i couldn't work out what rule would influence that. Fixed up manually

AlexHarker:
If I run clang format locally I get different results in relation to trailing comments. what version of clang format are you on (and are we working towards)? 

I believe that at least some of this is to do with AlignTrailingComments post clang format 16 (see docs). It is important to fix that as otherwise I can't easily test other changes.

olilarkin:
> Do we want to format 3rd party things like iPlug/Extras/HIIR/*?

I'll remove it. I want to update HIIR and move it to Dependencies

olilarkin:
> If I run clang format locally I get different results in relation to trailing comments. what version of clang format are you on (and are we working towards)?
> 
> I believe that at least some of this is to do with AlignTrailingComments post clang format 16 (see docs). It is important to fix that as otherwise I can't easily test other changes. I am on version 16.

yep... i had an old clang format, going to push updates...

olilarkin:
Blocked by https://github.com/iPlug2/iPlug2/issues/700 and https://github.com/iPlug2/iPlug2/pull/1168 (probably amongst others)

PR #954, opened 2023-04-18
Webview pr

hor-net:
some changes to the webview delegate to make it actually work in a plugin.


olilarkin:
thanks for this. here is the WIP branch for webview support, hopefully compatible with most of what you have done here.

https://github.com/iPlug2/iPlug2/tree/webview/rework-webview

olilarkin:
I am going to close this, since I have a lot of webview-reworking coming in on the webview/** branches that i believe cover it

Issue #955, opened 2023-04-27
makeRasterImage() needed in GPU as well as CPU when drawing Bitmap


sstillwell:
**CHECK**
Is this an issue or just a question? If it's just a question, then post on the [forum](https://iplug2.discourse.group), or ask on slack, don't pollute our issue tracker.
    This is an issue

**Describe the bug**
    When bitmaps are not stored as decompressed, rasterized images, performance takes a hit when controls are drawn frequently, as in bitmap knob stacks or bitmap meter displays.

**To Reproduce**
    Create a plugin using bitmap stacks on graphics controls, especially if they overlap.

**Expected behaviour**
    Performance should match that of any other vector control type.

**Screenshots**
    Not needed?

**IMPORTANT DETAILS**
* What plug-in format does it relate to - VST2, VST3, AUv2, AUv3, AAX, APP or ALL?
     Applies to all formats (haven't tested APP, though, I don't build for that)
* What platform does it relate to - Windows/macOS/iOS/Linux or ALL?
    Applies to Windows/Mac (haven't tested iOS or Linux, but imagine that this will affect all)
* What IGRAPHICS_BACKEND does it relate to, IGRAPHICS_NANOVG, IGRAPHICS_SKIA or ALL?
    Only applies to IGRAPHICS_SKIA.

**Additional context**
    None provided.

Pull request will be coming along shortly.

AlexHarker:
Can you supply some relevant bitmaps/a small example plug for testing? This seems sensible, but it'd be good to test properly before merging the PR.

sstillwell:
I don't have something offhand that I can send you, but if you profile any of the existing example plugins that use bitmap knobs backed by PNG graphics, you should see that an inordinate amount of time is spent uncompressing the image over and over with each new frame drawn.  @olilarkin had mentioned image cacheing on the GPU, but it doesn't look like that's happening here?  Maybe if the graphics are stored differently (i.e. compiled binary resources), but when stored as plain PNGs this happens.

olilarkin:
It would be interesting to try this again with the recent skia update, also with #1151

PR #956, opened 2023-04-27
makeRasterImage() needed whether CPU or GPU rendering

sstillwell:
There appears to be an especially-large performance hit on GPU if controls overlap on screen.

This PR fixes problem described in #955 
Performance is impacted when drawing bitmaps, especially controls with bitmap stacks, and even more especially if controls overlap (with associated extra draws/redraws)
This is the smallest possible PR to address this bug - we literally just stop conditionally rasterizing only for CPU rendering.

thanks, we like PRs!

oli & alex

### DIFF
--- a/IGraphics/Drawing/IGraphicsSkia.cpp
+++ b/IGraphics/Drawing/IGraphicsSkia.cpp
@@ -78,9 +78,7 @@ IGraphicsSkia::Bitmap::Bitmap(const char* path, double sourceScale)
   
   auto image = SkImage::MakeFromEncoded(data);
   
-#ifdef IGRAPHICS_CPU
   image = image->makeRasterImage();
-#endif
   
   mDrawable.mImage = image;
   
@@ -93,9 +91,7 @@ IGraphicsSkia::Bitmap::Bitmap(const void* pData, int size, double sourceScale)
   auto data = SkData::MakeWithoutCopy(pData, size);
   auto image = SkImage::MakeFromEncoded(data);
   
-#ifdef IGRAPHICS_CPU
   image = image->makeRasterImage();
-#endif
   
   mDrawable.mImage = image;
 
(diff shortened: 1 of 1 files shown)


sstillwell:
I'm not sure why all this would have failed...it looks like code download failed in a couple of places, so maybe build failed due to missing files?  Certainly the change in the PR is trivial and is working successfully in my local GitLab repo.

olilarkin:
So the reason I haven't merged this yet is because I would have thought that keeping bitmaps on the GPU would be a performance win. Need to investigate more

AlexHarker:
I am looking at this now and I don't see any difference on IGraphicsStressTest for either Skia CPU or Metal if I turn this on or off.

I'm using the DrawFittedBitmap test which I've modified to make the control dirty every frame. 
On my Mac I get (regardless of the use of rasterImage():

NanoVG Metal - 4000 items drops to about 50fps
Skia Metal - 2000 items drops to about 45 fps
Skia CPU - 60 items drops to about 42 fps

@sstillwell - did these performance things crop up on windows or Mac or both?

AlexHarker:
Additionally we are now using this:

https://api.skia.org/namespaceSkImages.html#ad56f01c94c380cff09a444b374b2aa28

I suspect that this may make explicit rasterisation totally irrelevant. I will try to test on the old skia also.

AlexHarker:
With the old version of Skia I get slightly slower for CPU (38 fps) but again makeRasterImage() makes no difference.
I am on an M3 Pro - not sure if platform is a factor here, but yet to get to any meaningful comparisons of this call.

sstillwell:
I'm going on memory here, but I believe we were noticing it in profiling on macOS - the amount of time spent in decompressing PNGs every time an image was drawn - the big culprit for us were meters drawn from bitmap stacks, which a) made the image large, and 2) made it necessary to uncompressed just to be able to find the frame within the stack.

olilarkin:
Closing this and will reopen if it is reproducible with the new skia libs

Issue #962, opened 2023-05-15
Make AAX automation indicator default to target RECT rather than bounds RECT 


sstillwell:
When creating a control, the bounds of the control may include large amounts of extra space, say for shadows or reflections onto the background.  The current defaults would place the automation indicator possibly far from the nominal control, even overlapping other controls, leading to confusion as to what the automation indicator is actually pointing to.

A possible solution here (other than overriding the `DrawPTHighlight()` method) is to make the indicator take its position reference from the `mTargetRECT` rather than `mRECT`, allowing the developer to pull the indicator closer to the "business end" of the control.  This would result in no change by default, as mTargetRECT defaults to mRECT unless changed.

PR #963, opened 2023-05-15
Move AAX automation indicator to target RECT

sstillwell:
Bitmap controls may have larger bounds to accommodate shadows, reflections, etc. and placing the automation indicator at the bounds of the control can be confusing.

  See issue:#962
  Allows developer to move AAX automation indicator closer to the actual control when the bounds of the control are larger to accommodate shadows, etc. in bitmaps.
  This change is graphics back-end agnostic.

thanks, we like PRs!

oli & alex


olilarkin (approved):
LGTM

PR #966, opened 2023-05-17
Dismiss the view controller before handling the selection (see #965)

### LINKED ISSUE #965: IOS menu selection cannot trigger a text entry or other presentation from the view controller
**Describe the bug**
The IOS menu handling does not dismiss the view controller until after the selection is handled. Thus if yo

**To Reproduce**
Create a plugin with a UI in which a menu selection directly triggers CreateTextEntry. The text entry does not work and an error ensues because the view controller is already presenting.

**Expected behaviour**
There is a text entry triggered (which there is on desktop)

**IMPORTANT DETAILS**
* What plug-in format does it relate to - VST2, VST3, AUv2, AUv3, AAX, APP or ALL?
* What platform does it relate to - Windows/macOS/iOS/Linux or ALL?
* What IGRAPHICS_BACKEND does it relate to, IGRAPHICS_NANOVG, IGRAPHICS_SKIA or ALL?

Auv3 on IOS only (works on desktop)

AlexHarker:
This PR fixes #965 by dismissing the menu before calling the handler - it is simply a code reordering and should not have any negative side-effects

### DIFF
--- a/IGraphics/Platforms/IGraphicsIOS_view.mm
+++ b/IGraphics/Platforms/IGraphicsIOS_view.mm
@@ -160,14 +160,14 @@ extern StaticStorage<CoreTextFontDescriptor> sFontDescriptorCache;
 
   if (pItem->GetIsChoosable())
   {
+    [self dismissViewControllerAnimated:YES completion:nil];
+
     mMenu->SetChosenItemIdx(cellIndex);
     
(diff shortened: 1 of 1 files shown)


olilarkin (approved):
LGTM, tried it and it does what it say's on the tin

PR #969, opened 2023-05-18
Rework IOS text entry to allow cancelling (and with OK as default)

### LINKED ISSUE #968: IOS platform text entry cancel button doesn't cancel (still commits the edit)
**Describe the bug**
When entering values on IOS using the platform text entry the cancel button doesn't cancel and the the edit is still committed.

**To Reproduce**
Do as described above.

**Expected behaviour**
The edit should be cancelled and have no effect.

**IMPORTANT DETAILS**
* What plug-in format does it relate to - VST2, VST3, AUv2, AUv3, AAX, APP or ALL?
* What platform does it relate to - Windows/macOS/iOS/Linux or ALL?
* What IGRAPHICS_BACKEND does it relate to, IGRAPHICS_NANOVG, IGRAPHICS_SKIA or ALL?

IOS AUv3 only.

AlexHarker:
This PR does three things:

1 - fixes #968 (so that cancel works) 
2 - puts the OK button on the right (and cancel on the left)
3 - sets OK as default (highlighted)x

### DIFF
--- a/IGraphics/Platforms/IGraphicsIOS_view.mm
+++ b/IGraphics/Platforms/IGraphicsIOS_view.mm
@@ -392,17 +392,6 @@ extern StaticStorage<CoreTextFontDescriptor> sFontDescriptorCache;
   mMTLLayer = nil;
 }
 
-- (void) textFieldDidEndEditing:(UITextField*) textField reason:(UITextFieldDidEndEditingReason) reason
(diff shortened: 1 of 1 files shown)


olilarkin (approved):
LGTM, but I'd squash the two commits when merging

PR #970, opened 2023-05-18
Nihoteph iplug2 warnings and tab

nihoteph:
Fixes various cast warnings double->float and double->int

Adds a SelectPage to the VTabbedPagesControl
Checks for presence of bubbleControl in VTabbedPagesControl to avoid crashing if null.

### DIFF
--- a/Examples/IPlugControls/IPlugControls.cpp
+++ b/Examples/IPlugControls/IPlugControls.cpp
@@ -3,6 +3,11 @@
 #include "IPlugPaths.h"
 #include "IconsForkAwesome.h"
 #include "IconsFontaudio.h"
+#include "PageMiscControls.h"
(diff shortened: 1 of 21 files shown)


olilarkin:
thanks for the PR :-)

BUT...

A good PR should change one thing. If you want to add something to the WIP tabbed panel stuff, then you need to make a PR to that branch not master

If you want to make a PR for the float conversions, again that needs to be an atomic change.

You could either rebase and force push to you branch to fix this, or we can close the PR and make another branch & PR for each thing. as it stands this wont get merged

olilarkin:
also i already fixed the bubble control thing, thanks for spotting that

olilarkin:
Thanks, I have added all these changes on the controls/container branch

PR #990, opened 2023-06-27
`IDirBrowseControlBase::ScanDirectory` finds files whose names include the extension elsewhere in their path.

### LINKED ISSUE #989: `IDirBrowseControlBase::ScanDirectory` fails to find files whose names include the extension elsewhere in their path.
**Describe the bug**
This surfaces https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/306, whose root cause is with [`DirBrowseControlBase::ScanDirectory()`](https://github.com/iPlug2/iPlug2/blob/b3a51573f45c830755ee4426d2d1eb04cf645359/IGraphics/IControl.cpp#L1120), which does not function as I think it's intended to.

The Issue is that files whose path contains the string contained in `mExtension` will _not_ match if the extension occurs anywhere _before_ the file extension (earlier in the file _or_ in the name of a containing directory).

**To Reproduce**
* Run NAM version 0.7.4
* Select a file in a directory containing the following files:
  * `5150.nam`
  * `birdie nam.nam`
  * `DR Pedal Platform.nam`
  * `Fender 65 Twin.nam`

**Expected behaviour**
All four files listed above should be pushed to `mFiles`.

**Screenshots**
See linked issue above.

**IMPORTANT DETAILS**
* Not specific to a plug-in format.
* Not specific to a platform.
* Not specific to a graphics backend.

**Additional context**
The issue can be resolved by replacing the [use of `strstr` here](https://github.com/iPlug2/iPlug2/blob/b3a51573f45c830755ee4426d2d1eb04cf645359/IGraphics/IControl.cpp#L1141) with `strrstr`, defined as follows (I'm not aware of a built-in alternative):

```c
// Find the last occurrence of str2 in str1.
// Return a pointer to the first character of the match.
// Return a pointer to the start of str1 if str2 is empty.
// Return a nullptr if str2 isn't found in str1.
const char* strrstr(const char* str1, const char* str2) {
  if (*str2 == '\0')
    return str1;
  
  const char* result = nullptr;
  
  while (*str1 != '\0') {
    if (std::strncmp(str1, str2, std::strlen(str2)) == 0)
      result = str1;
    
    str1++;
  }
  
  return result;
}
```
Testing this appears to resolve the problem in the downstream plug-in:

image

sdatkinson:
Define `strrstr`, which behaves like `strstr` but finds the _last_ instance instead of the first.
Use it in ScanDirectory()

Resolves #989.

PR Checklist:
* ✅ [create an issue](https://github.com/iPlug2/iPlug2/issues/new/choose) first to outline the problem, to reference in this PR
* ✅ clearly describe the problem that the PR fixes
* ✅ each PR should address one thing. It can include multiple commits, but you should try and tidy up work done into logical units if possible.
* ✅ pay attention to [iplug2 coding style](https://github.com/iPlug2/iPlug2/blob/master/Documentation/codingstyle.md)
  * _Mimic style of `strstr` for new function `strrstr`_
* ✅ If it applies to an option e.g. IGRAPHICS_NANOVG or VST3 plug-ins, try to provide fixes for all options (e.g. all graphics backends or all plug-in formats)
  * _N/A_

### DIFF
--- a/IGraphics/IControl.cpp
+++ b/IGraphics/IControl.cpp
@@ -1117,6 +1117,27 @@ void IDirBrowseControlBase::CheckSelectedItem()
   }
 }
 
+// Find the last occurrence of str2 in str1.
+// Return a pointer to the first character of the match.
+// Return a pointer to the start of str1 if str2 is empty.
+// Return a nullptr if str2 isn't found in str1.
+const char* strrstr(const char* str1, const char* str2) {
+  if (*str2 == '\0')
+    return str1;
+  
+  const char* result = nullptr;
+  
+  while (*str1 != '\0')
+  {
+    if (std::strncmp(str1, str2, std::strlen(str2)) == 0)
+      result = str1;
+    
+    str1++;
+  }
+  
+  return result;
+}
+
 void IDirBrowseControlBase::ScanDirectory(const char* path, IPopupMenu& menuToAddTo)
 {
   WDL_DirScan d;
@@ -1138,7 +1159,7 @@ void IDirBrowseControlBase::ScanDirectory(const char* path, IPopupMenu& menuToAd
         }
         else
         {
-          const char* a = strstr(f, mExtension.Get());
+          const char* a = strrstr(f, mExtension.Get());
           if (a && a > f && strlen(a) == strlen(mExtension.Get()))
           {
             WDL_String menuEntry {f};


sdatkinson:
Hey @olilarkin! Let me know if there's anything you need on this one.

olilarkin:
addressed in [b8473be](https://github.com/iPlug2/iPlug2/commit/b8473bea0c2df54260ea430b361e7254071bfd08)

PR #1007, opened 2023-10-04
Update EditorConfig to reduce noise in commits due to whitespace changes

tristan00b:
Addresses the config part of [issue 1008 - Fixing whitespace violations](1008)

Changes:
- `.editorconfig` 
  - Preamble updated to reflect current usage
  - Simplified catch-all glob expression
  - Set `insert_final_newline = true`
  - Set `end_of_line = crlf` for Windows shell scripts
- `.gitattributes`
  -  Set rules for CMake and Powershell files to `lf` and `crlf`, respectively
 
---

thanks, we like PRs!

oli & alex

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,14 +1,14 @@
-# top-most EditorConfig for IPlugQuake
+# Top-most EditorConfig for iPlug2
 root = true
 
-[*.*]
-
+[*]
 end_of_line = lf
-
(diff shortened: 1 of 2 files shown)


tristan00b:
Changed my mind, I think I'd like to reformulate the issue, which is *dealing with whitespace violations*. I will create a formal issue and propose my solution.

PR #1010, opened 2023-10-07
Clean up IPlugOSC

olilarkin:
The IPlugOSC classes are full of /** todo */ comments and expose a lot of private code. This PR goes some way to improving that, although more work can be done.


AlexHarker:
There is still a static tmp WDL_HeapBuf involved, which looks likely to be an issue. Other than that all static variables (members and in methods) are gone except the one which is used to store a const set of characters (hdr[16]) which seems fine.

IPlug/Extras/OSC/IPlugOSC_internal.cpp
```diff
@@ -0,0 +1,413 @@
...
+  for (auto i = 0; i < nDevices; i++)
+  {
+    auto* pDev = gDevices.Get(i);
+    if (pDev->mHasInput)
+      pDev->RunInput();
+  }
+
+  if (mIncomingEvents.GetSize())
+  {
+    static WDL_HeapBuf tmp;
```
AlexHarker, line 237: Here is the static in question

PR #1014, opened 2023-10-21
IOS make sure that reused cells display correctly enabled

### LINKED ISSUE #1013: IOS native popups disabled items act incorrectly
**Describe the bug**
When using native popup menus on IOS that have disabled items some arbitrary menu items are displayed as disabled when this has not been requested

**To Reproduce**
Build and IOS app with the native popup and a menu with some disabled items.

**Expected behaviour**
Only items marked as disabled are greyed out.

**IMPORTANT DETAILS**
* What plug-in format does it relate to - AUv3
* What platform does it relate to - iOS
* What IGRAPHICS_BACKEND does it relate to, ALL?

AlexHarker:
This fixes #1013

The source of the bug is the reuse of the same cellblock, and previously only setting the relevant properties for disabled items. This PR makes sure that properties are set for every call so that the disabled/enabled items are correct.

### DIFF
--- a/IGraphics/Platforms/IGraphicsIOS_view.mm
+++ b/IGraphics/Platforms/IGraphicsIOS_view.mm
@@ -128,6 +128,11 @@ extern StaticStorage<CoreTextFontDescriptor> sFontDescriptorCache;
     cell.userInteractionEnabled = NO;
     cell.textLabel.enabled = NO;
   }
+  else
+  {
+    cell.userInteractionEnabled = YES;
+    cell.textLabel.enabled = YES;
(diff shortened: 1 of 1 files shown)


olilarkin:
Makes sense, although i can't reproduce the issue in a quick test. 

could do if those objc YES NO are convertible from bool
```
cell.textLabel.enabled = cell.userInteractionEnabled = pItem->GetEnabled();
```

olilarkin (approved):
LGTM, i pushed a fixup to make it a one liner

Issue #1024, opened 2023-11-27
handle multiple files being drag-n-dropped


svantana:
IGraphics::OnDrop() only reports the first file if multiple files are dropped.

PR #1025, opened 2023-11-27
handle multiple items being drag-and-dropped: OnDropMultiple

svantana:
Adds interface to handle multiple items being drag-and-dropped: OnDropMultiple(). Falls back to single file version if not implemented.

solves issue: https://github.com/iPlug2/iPlug2/issues/1024

implemented on mac and windows as those are the platform that currently supports drag-n-drop.


olilarkin:
Nice: will look at trying this and merging it v soon. On first glance, looking good

olilarkin:
Thanks again, i tweaked a bit and added it in #1030

PR #1037, opened 2023-12-07
Fix LanczosResampler's magic `4`s

### LINKED ISSUE #1036: Lanczos resampler uses hard-coded `4`s instead of appropriate varaible `A`
**Describe the bug**
The use of `4` on lines 206, 207, 219, 226, and 227 of [`LanczosResampler.h`](https://github.com/iPlug2/iPlug2/blob/2376d58e1d157c6b8f9b1bd3ac673bcaa5651649/IPlug/Extras/LanczosResampler.h#L70) appears incorrect. Instead, they should use [`A`](https://github.com/iPlug2/iPlug2/blob/2376d58e1d157c6b8f9b1bd3ac673bcaa5651649/IPlug/Extras/LanczosResampler.h#L73).

Changing `A` as-is (to e.g. `8`) results in very incorrect-sounding behavior!

**To Reproduce**

* Change `A` to `8`.
* Compile something using the `NonIntegerResampler`
* Bounce something at both native sample rate and something else.
* Something terribly wrong like this ensues:

image

---

I'll have a PR along presently.

Fixing this bug as well as cranking `A` up to 12 fixes both https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/398 and  https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/400. However, I think that `A` should be chosen based on the needs of its use case, so I'm going to leave it at `4`. [Ideally, I wish it weren't hard-coded, but that's a bit more non-trivial due to the use of some static variables that bake in the assumption that it's not getting messed with a bit too hard for a quick fix.]

sdatkinson:
Replace magic `4`s with `A`.
Resolves #1036.

### DIFF
--- a/IPlug/Extras/LanczosResampler.h
+++ b/IPlug/Extras/LanczosResampler.h
@@ -203,8 +203,8 @@ private:
 
     f0 = _mm_add_ps(f0, _mm_mul_ps(df0, fl));
 
-    auto f1 = _mm_load_ps(&sTable[tidx][4]);
-    auto df1 = _mm_load_ps(&sDeltaTable[tidx][4]);
+    auto f1 = _mm_load_ps(&sTable[tidx][A]);
+    auto df1 = _mm_load_ps(&sDeltaTable[tidx][A]);
     f1 = _mm_add_ps(f1, _mm_mul_ps(df1, fl));
 
     for (auto c=0; c<NCHANS;c++)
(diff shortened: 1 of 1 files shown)


olilarkin:
Thanks... it required a bit more work... in https://github.com/iPlug2/iPlug2/commit/4bfc3e15d67f0c7c930aab71d1bf0e551362a1c4

Issue #1038, opened 2023-12-16
`RealtimeResampler` has errors


sdatkinson:
I've gotten some reports since shipping the resampler in https://github.com/sdatkinson/NeuralAmpModelerPlugin/releases/tag/v0.7.7 of a few buggy behaviors:

* When resampling from a DAW sample rate of 44.1k to a rendering sample rate of 48k, there's an oddly large latency--far more than the (approximately) `2*A` that should be required by the Lanczos resampler being used twice (up/down).
* The latency depends on the buffer size--something like 150 at a buffer size of 128, and 1150 at a buffer size of 1024.

I did some digging and I think I've figured out what's going on. The issues appear to be 

So if I've understood everything correctly, then here are the issues that cause the problem:
* [The latency calculation in `Reset()`](https://github.com/iPlug2/iPlug2/blob/40ebb560eba68f096221e99ef0ae826611fc2bda/IPlug/Extras/RealtimeResampler.h#L104) is wrong. Instead of `2 * mResamplerUp->GetNumSamplesRequiredFor(1)`, it should be `mResamplerUp->GetNumSamplesRequiredFor(mResamplerDown->GetNumSamplesRequiredFor(1))`.
* The [`while` loop](https://github.com/iPlug2/iPlug2/blob/40ebb560eba68f096221e99ef0ae826611fc2bda/IPlug/Extras/RealtimeResampler.h#L146) logic is wrong. Instead of `mResamplerUp->GetNumSamplesRequiredFor(outputLen) == 0` (i.e. only do resampling if you can make `outputLen` internal samples--which is the _larger_ value for the integer math!--it should be `mResamplerUp->GetNumSamplesRequiredFor(1) == 0`--run the code so long as there's anything for the first resampler to pop. (The contents of the while loop need to be ready to work with varying block sizes--which is expected for the integer math. The return values on the `PopBlock`s are also needed to be able to know how much you're working with on any given time throughout the function.
* Also a slightly minor issue, [`mScratchPtrs` is used for both the input and output to `func`](https://github.com/iPlug2/iPlug2/blob/40ebb560eba68f096221e99ef0ae826611fc2bda/IPlug/Extras/RealtimeResampler.h#L149C22-L149C22), which (invalidly) assumes that `func` can be done in-place. (Luckily, it's not a problem for my use case).
[...]

PR #1039, opened 2023-12-17
Rework RealtimeResampler

olilarkin:
This PR addresses several issues in #1038 and generally reworks the class based on @sdatkinson's findings

It also adds an example project to test the resampler and its latency reporting

- Clean up and clarify
- Remove cubic interpolation
- Resampling type is only set at compile time
- Fix added latency
- Latency reporting adjusted for null-signal


olilarkin:
NOTE: should also optimize it not to reallocate memory and reallocate the resamplers in Reset() if the samplerate etc didn't change

olilarkin:
merged, without example

Issue #707, opened 2021-03-16
Mac, ITextEntryControl.cpp: Minus key not recognized


TBProAudio:
Minus key of Mac(Book) keyboard (key close to enter key) is not recognized. This means the user cannot enter negative numbers. 

To Reproduce:
Use any control with input prompt, enable global mTextEntryControl by using pGraphics->AttachTextEntryControl();

Possible fix:

add 2 times

```
#if defined(OS_MAC)
            if (key.VK == kVK_INSERT)                       
            {
                stbKey = '-';
                break;
            }
#endif
```
to switch (pParam->Type()){}, kTypeInt and kTypeDouble

IMPORTANT DETAILS

    What plug-in format does it relate to - vst2/vst3/aax/au
    What platform does it relate to - Mac

AlexHarker:
I can't reproduce this issue here - my minus key picks up fine - can you give more details about your keyboard and keyboard configuration? Is this the only minus sign on your keyboard, or is it a secondary one?

AlexHarker:
I get the correct stbKey of 45 ("-") with my keyboard so there's definitely an issue with the key detection on your setup here - we'd need more details to investigate further.

TBProAudio:
Maybe it is an issue with the keyboard/locale layout/settings in IPLUG2:

I have a German layout: https://keyshorts.com/blogs/blog/37615873-how-to-identify-macbook-keyboard-localization#german
which has the "minus" key close to the right shift key.
So with German layout I get kVK_INSERT when minus is pressed...

1) The issue is Mac only
2) App works, VST2/VST2/AAX and AU not

maybe this helps.

AlexHarker:
OK - I think we are probably looking in the wrong place - would you be able to tell me what rawcode value you get for that key in MacKeyEventToVK in IGraphicsMac_view.mm (line 97 on master)? 

Is it 114? If so then some assumptions about virtual key mappings on Mac might be where the issue is.

TBProAudio:
Hmm, IGraphicsMac_view.mm breaks only with App.
Does not break with vst2.

AlexHarker:
My bad - I misread your original report - the issue here is what we are getting passed from VST2VST3 key handlers - I CAN'T see a reason why AU and AAX wouldn't work as these should go through igraphics view - can you test with audio unit?

I can now reproduce on VST2 / VST3 in reaper and we'll need to modify the key codes appropriately somewhere.

AlexHarker:
[...]

PR #1042, opened 2023-12-27
Incorrect VST key press events in reaper

olilarkin:
As noted in #707 certain keypresses are getting garbled in Reaper when they are handled via the VST mechanisms. They are handled by that in the first place rather than IGraphics, so that we can for instance catch the space bar in a text entry control before it is handled by the host to start the transport. This is particularly an issue for typing a - sign in an ITextEntryControl. This PR fixes it by only using the Steinberg IPlugView::onKeyDown/Up() overrides for the SPACE bar in Reaper, since that is something we want to catch using the VST methods because it will be blocked from IGraphics, we use IGraphics for other keys in Reaper. The same problem is there in VST2 with reaper, work around that too.


olilarkin:
Will merge this and address once the bug is fixed in Reaper

PR #1043, opened 2024-01-01
App/separate settings dialog

AlexHarker:
This is work towards the 

It is based on https://github.com/iPlug2/iPlug2/tree/app/flexible-io

The work is the following:

- iPlugAPP clean-up and spacing changes
- RTAudio update to 6.0.1 (and changes to fix builds with this version)
- Splitting of the settings dialog into a separate file/class in a way that allows different approaches to this window (last commit only)

Notes:

- -1 is not used as a special value for audio device IDs as RTAudio does not guarantee what values IDs might take. ValidatedID is used where it is needed to indicate if the ID is empty
- Renaming has been used to clarify class members and types (e.g. State -> Settings)
- IPlugAppDialog is included in IPlugAppHost by composition and various methods have been made private to more fully separate the settings dialog from the host
- MainDlgProc is in the host file, but the settings dialog has its own DlgProc
- *none of the changes thus far change the operation of the settings dialog or the format of the saved data. Merging should not break any functionality.


olilarkin:
Gave it a quick look through and compiled, looks good so far

PR #1056, opened 2024-02-04
iOS App: Restart audio engine when configuration changes

### LINKED ISSUE #987: iOS app freezes when connecting an external audio interface, or headphones
need to implement handing of the correct notification and restart the AVAudioSession

https://developer.apple.com/documentation/avfaudio/avaudiosessionroutechangenotification?language=objc

olilarkin:
This PR should fix #987. More work is required to check that e.g. changing samplerates and channel counts propagate correctly, and may require rebuilding the AVAudioEngine node graph


olilarkin:
LGTM oli

PR #1057, opened 2024-02-04
Ios/midi io

olilarkin:

thanks, we like PRs!

oli & alex

### DIFF
--- a/IPlug/AUv3/iOSApp/IPlugIOSMIDI.mm
+++ b/IPlug/AUv3/iOSApp/IPlugIOSMIDI.mm
@@ -95,14 +95,14 @@ long IPlugIOSMIDI::GetIndexFromName(WDL_String name, ConnectionType type)
 //static
 void IPlugIOSMIDI::SetMidiPort(const char *name, ConnectionType type)
 {
-  NSDictionary* dic = @{@"name": @(name), @"direction": @(type == ConnectionType::Source ? "Source" : "Destination")};
+  NSDictionary* dic = @{@"name": @(name), @"direction": @(type == ConnectionType::Source ? "source" : "destination")};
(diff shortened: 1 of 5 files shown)


IPlug/AUv3/iOSApp/IPlugIOSMIDI.mm
```diff
@@ -0,0 +1,230 @@
...
+void IPlugIOSMIDI::SetMidiPort(const char *name, ERoute route)
+{
+  NSDictionary* dic = @{@"name": @(name), @"direction": @(route == iplug::kInput ? "source" : "destination")};
+  [[NSNotificationCenter defaultCenter] postNotificationName:@"SetMIDIPort" object:nil userInfo:dic];
+}
+
+//static
+void IPlugIOSMIDI::GetMidiPort(WDL_String& name, ERoute route)
+{
+  NSDictionary* dic = @{@"name": [NSValue valueWithPointer:&name], @"direction": @(route == iplug::kInput ? "source" : "destination")};
```
olilarkin, line 114: Storing a pointer to the WDL_String here and setting the value in the notification handler feels wrong here... trying to think of other options @AlexHarker

AlexHarker: Yes - I can see why that's not ideal, but we are picking it up at the other end and the idea was to be able to get the value out here - I wouldn't want to pass a pointer to an plugin-developer in this way, but here the framework is using the notification system to pass around data. As far as I am aware (and I am pretty sure I would have tested before using this approach) the notification system is immediate and synchronous, so it should be safe - it's just that the notification system is the best way to communicate between app and plugin (given that the framework has no notion of the wrapping app). This was based on the BT Midi notification stuff - if there's a better way or a way that involves putting a hook in the plugin base etc. then I'd be ok with that.

PR #1058, opened 2024-02-04
Make GetWantsGestures() virtual

AlexHarker:
As discussed on call - this allows controls to accept gestures in a custom manner.

This has two advantages:

1 - for users who want to still accept gestures whilst an animation is running (currently disallowed)
2 - for users who are bypassing IControl::AttachGestureRecognizer(), attaching directly to IGraphics, and then overriding IConrol::OnGesture()

Both of these seem useful, and there is no effect on previously written code by making this virtual.


olilarkin (approved):
LGTM

Issue #1060, opened 2024-02-11
MacOS OpenGLContext leaks ~4MB on window close


svantana:
OpenGL resources are init'ed on plugin window open but not released on close, resulting in a memory leak. This happens with both NanoVG and Skia backends, which strongly suggests that the error is in iPlug2.

How to verify:
Open and close the iPlugEffect editor in any mac host 50 times, and note in activity monitor that the host's memory has gone up by 200MB.

olilarkin:
thanks!

PR #1061, opened 2024-02-11
Fixed memory leak when closing opengl window on macos (fix for issue #1060)

svantana:
Releasing these resources fixes the leak.


olilarkin:
nice work. tried it here and it works

olilarkin (approved):
LGTM

Issue #1062, opened 2024-02-12
WebUI example crashes when double-clicked


Vector-Hector:
**Describe the bug**
When double clicking the built WebUI example VST3 plugin, FL Studio will crash. 

**To Reproduce**
- Install the latest IPlug2 as described in the [Getting Started](https://github.com/iPlug2/iPlug2/wiki/02_Getting_started_windows)
- Build the IPlug2WebUI example
- Make sure FL Studio finds the plugin
- Add it to an FX slot
- Make sure the plugin is currently closed
- Double click the plugin

**Expected behaviour**
I expected it to just open and close very quickly.

**IMPORTANT DETAILS**
* What plug-in format does it relate to - Currently only tested on VST3
* What platform does it relate to - Currently only tested on Windows 11 with FL Studio 21
* What IGRAPHICS_BACKEND does it relate to, IGRAPHICS_NANOVG, IGRAPHICS_SKIA or ALL? - Not sure, I am pretty new. If there is a default in the WebUI example, then that. But I can look it up if you tell me how to.

**Additional context**
- This does not happen on non-WebUI examples as far as I have seen
- It also happens if the index.html can't be found, like by setting an invalid path

Vector-Hector:
Adding this to the IWebView::OpenWebView callback of CreateCoreWebView2Controller fixed it for me:
```cpp
if (mWebViewWnd == nullptr) {
  return S_OK;
}
```

Vector-Hector:
Not sure whether you accept pull requests / how you want them to be structured, so I won't do one.

AlexHarker:
We do accept PRs - ideally after an issue, so please submit one - we wouldn't consider this complete until it was fixed in the main repo so I'm reopening this - I hope that's OK.

PR #1063, opened 2024-02-15
WebView: crash when rapidly closing/reopening window in FLStudio on windows

Vector-Hector:
fixes https://github.com/iPlug2/iPlug2/issues/1062

I am still new, so this is a bit of a pragmatic fix. As far as I understand the issue I was having, when opening the plugin in FL Studio, this loading code is executed. Buf if the plugin is closed very quickly, this callback seems to get called anyway. So without some kind of check whether the state is still valid, this function may run into problems. And the check I came up with was 

```cpp
if (mWebViewWnd == nullptr) {
  return S_OK;
}
```

I couldn't test this code on any machine except my dev machine (Windows 11, VST3, FL Studio 21), but I don't see how it could cause problems on other machines.

Hope it helps, love your library <3


olilarkin:
thanks for the PR... currently working on webview stuff and i'll have a look at this soon

olilarkin:
Thanks, merged in c9db1db6f734c358332086c11c489456ab5ae97c

Issue #927, opened 2023-01-31
Implement external drag


olilarkin:
We should be able to initiate external drag and drop, for instance so a plug-in can drag midi files onto the DAW timeline.

PR #1064, opened 2024-02-21
external drag-n-drop

svantana:
solves issue #927

Tested on macOS 13.5.1 and Windows 10


olilarkin:
Great work, thanks for this! Did e.g. the windows implementation come from some other project? if so what was the license of that code? Going to clean this PR up a bit here #1067 in order to get the code in iPlug2 style etc.

svantana:
Right, I forgot about the two-space tabs etc, sorry! It's my own implementation, mostly just pretty bare bones derivations of the base classes.

Issue #1070, opened 2024-03-10
ShowMessageBox mixes up message and caption on mac


svantana:
IGraphicsMac.cpp, line 299:

the string 'msg' is put as the first argument of NSRunAlertPanel, which is called "title". Looking at the resulting messagebox,  that first argument is used as a headline, while the 5th argument 'cap' is put as a text paragraph below. Comparing with the same function on windows, it seems to me that these two should switch position.

svantana:
Sure, this code:
`G.ShowMessageBox("str", "caption", kMB_OK);`
looks like this on windows and mac respectively. It seems to me that windows is what one would expect, and mac is reversed.
<img width="111" alt="messageBox_win" src="https://github.com/iPlug2/iPlug2/assets/1098215/bd88cd40-e6d3-4084-b53a-29341f927c2c">
<img width="271" alt="messageBox_mac" src="https://github.com/iPlug2/iPlug2/assets/1098215/aa0d42ff-fe64-4e11-b3ab-e7de2e37cce6">

PR #1071, opened 2024-03-10
ShowMessageBox: fix for switched caption and string on mac

svantana:
Fix for issue #1070

The following code:
`G.ShowMessageBox("str", "caption", kMB_OK);`
before commit:
messageBox_mac
after commit:
messageBox_mac_fix

### DIFF
--- a/IGraphics/Platforms/IGraphicsMac.mm
+++ b/IGraphics/Platforms/IGraphicsMac.mm
@@ -296,23 +296,23 @@ EMsgBoxResult IGraphicsMac::ShowMessageBox(const char* str, const char* caption,
   switch (type)
   {
     case kMB_OK:
-      NSRunAlertPanel(msg, @"%@", @"OK", @"", @"", cap);
+      NSRunAlertPanel(cap, @"%@", @"OK", @"", @"", msg);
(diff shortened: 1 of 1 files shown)


svantana:
@AlexHarker thoughts?

AlexHarker:
Looks like this is probably more consistent and changes seem sensible. @olilarkin ?

olilarkin:
yes, this seems better. I am going to rename the argument "caption" -> "title", make the same change for iOS and label it BREAKING, since people will need to adapt their code

olilarkin:
yes, this seems better. I am going to rename the argument "caption" -> "title", make the same change for iOS and label it BREAKING, since people will need to adapt their code

olilarkin:
yes, this seems better. I am going to rename the argument "caption" -> "title", make the same change for iOS and label it BREAKING, since people will need to adapt their code

svantana:
sounds good!

olilarkin:
In fact i am going to update this code using NSAlert while i am at it

olilarkin:
Closing this PR to replace with #1074 , thanks for the contribution... i'll extend it a bit

Issue #1080, opened 2024-04-03
Macos OpenGL: multiple instances clash


svantana:
When using NanoVG+OpenGL on mac, multiple instances can clash, like in the following video:

https://github.com/iPlug2/iPlug2/assets/1098215/65762e57-91bb-4ad4-b26f-252df0e529a4

The apple docs are abysmal but from discussions it sounds like it depends on whether the instances are on the same thread or not.

https://stackoverflow.com/questions/2952156/determing-if-an-nsopenglview-has-a-valid-context-cocoa-osx

https://stackoverflow.com/questions/21798018/nsopenglcontext-destroying-textures-of-another-context

PR #1081, opened 2024-04-03
mac-opengl render: proper init and exit (fixes issue #1080)

svantana:
The clearCurrentContext solves issue #1080. It is also apparently an accepted solution to various mac+opengl issues on SO, for example:

https://stackoverflow.com/questions/21798018/nsopenglcontext-destroying-textures-of-another-context

Most UI repos out there seem to start render calls with makeCurrentContext and end with clearCurrentContext. That is what I had to go on since Apple's docs don't say why/when it should be used.

The addObserver solves an issue in FL Studio-mac where if a complex iPlug2 gui is opened and closed really quickly, it would crash the FL Studio renderer. Same here, no docs do go on, but it looks like something people have learned is good to do.


olilarkin:
superseded by #1097

Issue #1082, opened 2024-04-15
Dropping const qualifier in IGraphics/Platforms/IGraphicsWin


MeijisIrlnd:
<h3> Description </h3>
(I feel weird opening an issue about this because it's such a tiny issue, but was going to open a PR and it told me to, so sorry if this is overkill)<br>
Under MSVC and C++20, compilation (all targets) of Examples/IPlugEffect fails with 

```Conversion from string literal loses const qualifier (see /Zc:strictStrings)```

on the declaration of `wndClassName`. I assume something must have changed with assigning string literals to pointers in c++20 or something along those lines (?)

<h3> To Reproduce </h3>
I encountered it using the recently-rebased cmake branch, building IPlugEffect, and changing the c++ standard from 17 to 20 in the CMakeLists: 

```
iplug_target_add(_base INTERFACE
  INCLUDE ${PROJECT_DIR} ${PROJECT_DIR}/resources
  LINK iPlug2_NanoVG_GL2
  FEATURE cxx_std_20)
 ```

Other than this, C++20 with MSVC seems to work out the box!

PR #1083, opened 2024-04-15
fixed dropped const in IGraphicsWin under c++20 (MSVC)

MeijisIrlnd:
tiny (tiny and I mean absolutely miniscule) fix to get iPlug2, MSVC and C++20 all living in harmony ( ™️ )
I opened an [issue](https://github.com/iPlug2/iPlug2/issues/1082#issue-2242570652) reporting it, but essentially C++20 must have changed something about assigning string literals to non-const pointers, the fix was just to make `wndClassName` const. Hope it's at all helpful!


olilarkin:
thanks! merged

Issue #1084, opened 2024-04-19
[CMake branch] Clang support on windows


MeijisIrlnd:
Hiya, when trying to use a Clang toolchain on Windows on the cmake branch, compilation fails due to `/Zi` being present, which under an msvc-like frontend, generates pdb files. 
I had a bit of a rummage around, and found that `/Zi` is hardcoded in the current `iplug_configure_target` implementation - I was able to get it working by checking if the compiler's cli is msvc-like or not, and if not, replacing `/Zi` with `-DEBUG` in clang's case. 
Not sure if this is a viable solution for `g++`, but it works with Clang on windows, I'll link the PR with the fix to this issue,
Hope it's at-all useful!

olilarkin:
thanks for this. I've added it to the CMake branch WIP

PR #1085, opened 2024-04-19
compiling with regular clang on windows

MeijisIrlnd:
Enables compiling under clang (rather than clang-cl) on Windows - described in a bit more detail in [this](https://github.com/iPlug2/iPlug2/issues/1084) issue, and I threw some comments around the bits I added for clarity. As mentioned in the issue, I haven't tried it with g++ under mingw or anything like that, but yeah, works for clang!


olilarkin:
thanks for this. I've added it to the CMake branch WIP

Issue #1086, opened 2024-04-20
CMakeLists for Examples/IPlugWebUI


MeijisIrlnd:
Hiya, unless I've made a very embarrassing oversight and it's on another branch (if it is don't tell me I'm not sure I can take it), there's no CMakeLists in the WebUI example. I've just finished setting it up on Windows, but I'm not enamoured with my solution (it involves relying on Nuget being *somewhere* on the user's path). I also haven't touched it on macOS yet, but I'm hoping to give that a go in the next few days, so in the meantime I'll open a draft PR and post here (and probably in the discord too) as things come up!

olilarkin:
thanks for this. I've added it to the CMake branch WIP

PR #1087, opened 2024-04-20
IPlugWebUI CMakeLists

MeijisIrlnd:
This is definitely a work in progress, and a little rough around the edges, but in its current form it compiles the WebUI example on Windows - macOS is on my todo list over the next few days, a few more thoughts in [this](https://github.com/iPlug2/iPlug2/issues/1086) feature request. Open to suggestions on other approaches here as well, let me know!

### DIFF
--- a/Documentation/cmake-docs.md
+++ b/Documentation/cmake-docs.md
@@ -1,0 +1,106 @@
+# CMake and iPlug2
+
+This documentation should help those trying to build iPlug2 projects using
+CMake. iPlug2 is a complex project and building it correctly on all supported
+platforms, with different compilers, IDEs, and plugin types complicates things
+even more.
+
+
+## Loading iPlug2's CMake scripts
+Loading iPlug2's CMake code is fairly easy, just a few lines.
+
+```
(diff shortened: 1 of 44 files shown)


MeijisIrlnd:
Managed to get it up and running on macOS - it runs the app, and builds au and vst3, but I haven't tried them in a host - I'll do that hopefully tomorrow, just need to jump through the codesigning hoops

MeijisIrlnd:
Okay works fine with AUV2 and VST3 as long as you sign them - I added some thoughts on this in the TAP Discord, but essentially I have a function I use for my JUCE stuff which will sign the target you're building, but I wasn't sure if you've already got facilities for this anywhere in the codebase, so err-ed on the side of leaving that out. Other things of note, not sure if I need extra flags for `cmake -G "Xcode"` in order for it to generate DSYMs, I'm not very familiar with Xcode specific things, and the NuGet stuff should probably be abstracted to a `FindWeb2.cmake` and `FindWIL.cmake`.
Let me know if you need me to add anything!

olilarkin:
thanks for this. I've added it to the CMake branch WIP

PR #1091, opened 2024-05-04
Add support for CLAP (CLever Audio Plugin API)

olilarkin:
This PR adds an iPlug2 API class for CLAP. This is almost all the hard work of @AlexHarker and has been made possible via support from U-He. A big clap/round of applause please! 👏 

https://u-he.com/community/clap
https://cleveraudio.org


IPlug/CLAP/IPlugCLAP.h
```diff
@@ -0,0 +1,250 @@
...
+  bool implementsRender() const noexcept override { return true; }
+  bool renderHasHardRealtimeRequirement() noexcept override { return false; }
+  bool renderSetMode(clap_plugin_render_mode mode) noexcept override;
+  
+  // clap_plugin_state
+  bool implementsState() const noexcept override { return true; }
+  bool stateSave(const clap_ostream *stream) noexcept override;
+  bool stateLoad(const clap_istream *stream) noexcept override;
+  
+  /*void stateMarkDirty() const noexcept {
```
olilarkin, line 131: @AlexHarker something unfinished here?

IPlug/CLAP/IPlugCLAP.h
```diff
@@ -0,0 +1,250 @@
...
+  
+  // clap_plugin
+  bool init() noexcept override;
+  bool activate(double sampleRate, uint32_t minFrameCount, uint32_t maxFrameCount) noexcept override;
+  void deactivate() noexcept override;
+  bool startProcessing() noexcept override { return true; }
+  void stopProcessing() noexcept override {}
+  clap_process_status process(const clap_process *process) noexcept override;
+  
+  //void onMainThread() noexcept override {}
```
olilarkin, line 111: @AlexHarker Also here?

AlexHarker:
This looks the same as the branch with the longer history to me, although there's a comment missing from IPlugVST3PRocessorBase.cpp (// TODO - push to GUI) which looks relevant if not CLAP related.

Happy for this to go in finally!

IPlug/CLAP/IPlugCLAP.h
```diff
@@ -0,0 +1,250 @@
...
+  };
+  
+public:
+  IPlugCLAP(const InstanceInfo& info, const Config& config);
+
+  // IPlugAPIBase
+  void BeginInformHostOfParamChange(int idx) override;
+  void InformHostOfParamChange(int idx, double normalizedValue) override;
+  void EndInformHostOfParamChange(int idx) override;
+//  void InformHostOfPresetChange() override;
```
olilarkin, line 92: TODO?

IPlug/IPlug_include_in_plug_src.h
```diff
@@ -270,6 +270,112 @@
...
+  #if PLUG_TYPE==0
+  #define CLAP_FEATURES CLAP_PLUGIN_FEATURE_AUDIO_EFFECT
+  #elif PLUG_TYPE==1
+  #define CLAP_FEATURES CLAP_PLUGIN_FEATURE_INSTRUMENT
+  #elif PLUG_TYPE==2
+  #define CLAP_FEATURES CLAP_PLUGIN_FEATURE_NOTE_EFFECT
+  #endif
+#endif
+
+std::string gPluginPath;
```
olilarkin, line 298: Are these globals going to be OK with multiple instances?

IPlug/CLAP/IPlugCLAP.h
AlexHarker: Do you mean the InformHostOfPresetChange()?

IPlug/CLAP/IPlugCLAP.h
AlexHarker: Iplug2 doesn't have mechanism to mark the state dirty,  but CLAP does - we could add it, but we'd need to discuss what is should be and how it should work.

IPlug/CLAP/IPlugCLAP.h
AlexHarker: This probably needs checking - the commented out things just mean there was a virtual in the base helper class that might need overriding, but I'm not sure what this method is supposed to do.

IPlug/IPlug_include_in_plug_src.h
AlexHarker: These globals don't relate to an instance - they are to do with entry to the shared library. In theory a host is allowed to load and unload without exact pairings, however, so I can take a look.

IPlug/CLAP/IPlugCLAP.cpp
```diff
@@ -0,0 +1,1015 @@
...
+{
+  IMidiMsg msg;
+  SysExData sysEx;
+  
+  // Transport Info
+  if (process->transport)
+  {
+    ITimeInfo timeInfo;
+
+    auto &transport = process->transport;
```
olilarkin, line 197: not immediately clear why auto& here?

AlexHarker: I can't see a reason - it's probably easier for the compiler to optimise a stack copy of the pointer, so this looks either like a mistake, or just that I was thinking about it weirdly when I did it.

IPlug/CLAP/IPlugCLAP.h
olilarkin: Yes, i removed it for now

IPlug/CLAP/IPlugCLAP.h
olilarkin: Removed for now

IPlug/CLAP/IPlugCLAP.h
olilarkin: DirtyParametersFromUI() needs to be reworked into DirtyState(). I've learnt recently that the VST3 implementation is problematic, will make an issue

IPlug/IPlug_include_in_plug_src.h
olilarkin: OK, seems to be fine

PR #1093, opened 2024-05-04
Add missing STDMETHODCALLTYPE macros

### LINKED ISSUE #1092: iPlug2 doesn't compile on windows in 32-bit due to missing macros
**Describe the bug**
iPlug2 doesn't compile on windows due to missing macros as part of IGraphics to do with drag and drop.

**To Reproduce**
Attempt to build a plugin with an IGraphics GUI on windows 32 bit.

**Expected behaviour**
It should compile.

**IMPORTANT DETAILS**
* What platform does it relate to - Windows
* What IGRAPHICS_BACKEND does it relate to - ALL?

AlexHarker:
This should fix #1092

### DIFF
--- a/IGraphics/Platforms/IGraphicsWin_dnd.h
+++ b/IGraphics/Platforms/IGraphicsWin_dnd.h
@@ -116,7 +116,7 @@ struct EnumFORMATETC final : public IEnumFORMATETC
     return E_NOINTERFACE;
   }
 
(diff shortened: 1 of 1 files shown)


AlexHarker:
Note that many of the methods were marked correctly with the macro, but those that are not cause windows compilation in 32bit to fail.

Issue #1094, opened 2024-05-07
AUv2 hosts reporting negative ppq becomes -1 in iPlug


svantana:
Logic, for example, does a little preroll with negative ppq when playing from bar 1. Also, I'm pretty sure I've seen other hosts where you can use time before bar 1, although I'm not sure which.

PR #1095, opened 2024-05-07
AUv2: allow negative ppq. Solves issue #1094

svantana:
VST3 uses the reported value even if negative, so I don't see why AUv2 should work any different.


olilarkin (approved):
LGTM

PR #1097, opened 2024-05-19
Activate and Deactivate GL context

### LINKED ISSUE #1080: Macos OpenGL: multiple instances clash

olilarkin:
Fixes #1080 and replaces #1081 

I noticed that there was also an issue when using the corner resizer with multiple iPlug2 windows open using IGRAPHICS_GL, this PR adapts #1081 to make sure that case is covered too.

@svantana would you be able to try it with the original bug and check if my modifications achieve the same fix as your original PR?

### DIFF
--- a/IGraphics/Drawing/IGraphicsNanoVG.cpp
+++ b/IGraphics/Drawing/IGraphicsNanoVG.cpp
@@ -480,6 +480,8 @@ void IGraphicsNanoVG::OnViewDestroyed()
 
 void IGraphicsNanoVG::DrawResize()
 {
+  ActivateGLContext();
+  
   if (mMainFrameBuffer != nullptr)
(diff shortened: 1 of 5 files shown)


svantana:
Yup seems to work. I tried both the multiple window issue and the move-to-external-display-with-other-resolution issue.

PR #1101, opened 2024-05-26
Add an option to allow items to be deleted from menus on IOS

AlexHarker:

This is a new feature for IPopUpMEnu in which items can be marked deletable, and these can be deleted (only on IOS) by using the normal IOS swipe and confirm method.

The menu is automatically updated, but the calling IControl also gets called to perform any underlying deletion work (for instance deleting a preset).


IGraphics/IGraphicsPopupMenu.h
```diff
@@ -95,6 +97,7 @@ class IPopupMenu
     
     void SetEnabled(bool state) { SetFlag(kDisabled, !state); }
     void SetChecked(bool state) { SetFlag(kChecked, state); }
+    void SetDeletable(bool state) { SetFlag(kDeletable, state); }
```
olilarkin, line 100: I added this setter

olilarkin (approved):
LGTM

PR #1109, opened 2024-07-11
Clearer wording for UnserializeEditorSize()

LudvigWard:
Mention editor size data instead of referencing parameter values in the doc comment, since the function is not meant to account for or work with parameter values


olilarkin (approved):
LGTM! Thanks

Issue #1113, opened 2024-07-23
IGraphics context doesn't know when the window looses focus


olilarkin:
When the user clicks away from the plug-in window, there should be a way for the IGraphics context to e.g. collapse the IPopupMenuControl, or cancel in progress edit of the ITextEntryControl. 

**IMPORTANT DETAILS**
* What plug-in format does it relate to - ALL
* What platform does it relate to - Windows/macOS & ??
* What IGRAPHICS_BACKEND does it relate to - ALL

**Additional context**
[Add any other context about the problem here.
](https://iplug2.discourse.group/t/how-to-catch-the-plugin-window-focus-loss-event-in-mac-os/920)

PR #1114, opened 2024-07-23
Add IGraphics::OnLostFocus

olilarkin:
Addresses #1113 1113

### DIFF
--- a/IGraphics/IGraphics.cpp
+++ b/IGraphics/IGraphics.cpp
@@ -1525,6 +1525,22 @@ void IGraphics::OnAppearanceChanged(EUIAppearance appearance)
     mAppearanceChangedFunc(appearance);
 }
 
+void IGraphics::OnLostFocus()
+{
+  ClearMouseOver();
+  ReleaseMouseCapture();
+  
+  if (mPopupControl && mPopupControl->GetExpanded())
+  {
+    mPopupControl->Hide(true);
+  }
+  
+  if (mTextEntryControl && mTextEntryControl->EditInProgress())
+  {
+    mTextEntryControl->Hide(true);
+  }
+}
+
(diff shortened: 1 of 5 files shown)


flutomax:
You also need to add the method to IControl:
```
/** This is an called when focus is lost. /todo check this */
  virtual void OnLostFocus() {}
```
And in class IGraphics -
to IGraphics.h:
`void OnLostFocus();`
to IGraphics.cpp:
```
void IGraphics::OnLostFocus()
{
  TRACE
  ForAllControls(&IControl::OnLostFocus);
}
```
In IGraphicsWin.cpp add WM_KILLFOCUS message processing like this:
```
case WM_KILLFOCUS:
    {
      pGraphics->OnLostFocus();
      return 0;
    }
```

olilarkin:
I don't think adding a method to IControl is how I will do it, since it's a global thing. I'll think about it.

flutomax:
> I don't think adding a method to IControl is how I will do it, since it's a global thing. I'll think about it.

How then will the final control receive notification of loss of focus? 
I suggested the simplest way - to add a new virtual method OnLostFocus(), which can be overridden in custom control classes. And this will not entail any changes in the current codes of both the framework and user files.

AlexHarker:
Is this just for losing focus on the window or more generally? Shouldn't this be rolled in with a future focus system that also deals with controls?

olilarkin:
It's a global thing for now, and I think I will add a callback like the IGraphicsKeyHandler. Adding the concept of focus for IControls is something bigger we can look at when we start the accessibility work

AlexHarker:
Cool - might I suggest something more like OnFocus() which gets called with a bool though as an alternative idea, or is this inline with how other platforms handle focus - it's not something I'm particularly familiar with.

PR #1151, opened 2024-11-20
Enable SKIA Graphite for Metal

olilarkin:
Skia's new Graphite backend is designed for modern graphics APIs like Metal. It should be faster and we should see less jank, e.g. brief pause on first render while shaders are compiled

To enable it, defined IGRAPHICS_SKIA and SKIA_GRAPHITE


olilarkin:
Drop shadow doesn't work
Screenshot 2024-11-20 at 23 22 47

AlexHarker:
Might be worth trying this in combination with this:

https://github.com/iPlug2/iPlug2/commit/6fd58c6d46101e2b428ec7d949f063f4ac1a6019

which implements a drop-shadow that should look basically identical with an image filter

Issue #508, opened 2020-04-26
Incorrect text size in IPlug2


Youlean:
You choose to adjust the size based on total height. That would make all fonts fit inside size, but it will make using multiple fonts inconsistent in size because different fonts will have different descended and ascender.

It should be tested if this isan issue or not.

Youlean:
It seems that there are 2 problems with fonts:

1. Font size
2. Font vertical position

**1. Problem** 
https://github.com/iPlug2/iPlug2/blob/546d024896c5b3faf098cef69ef15c1bf2e5b250/IGraphics/Drawing/IGraphicsCairo.cpp#L495

I don't think we should multiply font size by EM ratio. This will give us smaller fonts than what is standard, and it will make different fonts a different size.

Example with EM ratio correction:
<img width="828" alt="Screen Shot 2020-04-26 at 4 52 15 PM" src="https://user-images.githubusercontent.com/13148293/80311158-54000b80-87de-11ea-851e-33a27af1eb2b.png">

Example without EM correction:
<img width="829" alt="Screen Shot 2020-04-26 at 4 53 26 PM" src="https://user-images.githubusercontent.com/13148293/80311179-7d209c00-87de-11ea-8f6e-a0da4fa31b7f.png">

As you can see, without EM correction, we are getting more equal fonts in size.

**2. Problem**

The top and bottom positions should be fixed too.

Example of original positioning:
<img width="833" alt="Screen Shot 2020-04-26 at 4 26 10 PM" src="https://user-images.githubusercontent.com/13148293/80311225-bc4eed00-87de-11ea-8bac-b1cefd361d41.png">

Example for corrected positioning:
<img width="828" alt="Screen Shot 2020-04-26 at 4 46 32 PM" src="https://user-images.githubusercontent.com/13148293/80311248-df799c80-87de-11ea-8c04-c66aef8a4411.png">

These examples are drawn with Cairo.

Code for corrected positioning:

```
  switch (text.mVAlign)
  {
    case EVAlign::Top:      y = r.T + ascender - descender;                  break;
    case EVAlign::Middle:   y = r.MH() - descender + (textHeight / 2.0);     break;
    case EVAlign::Bottom:   y = r.B;                                         break;
  }
```

Youlean:
Skia vertical aligement:

```
  switch (text.mVAlign)
  {
    case EVAlign::Top:      y = r.T - ascender - descender;                break;
[...]

PR #1164, opened 2024-11-30
WIP: Breaking: Fix IGraphics Text Size, update NanoVG

olilarkin:
See discussion on the issue:

We should also think about the new text shaping facilities in Skia via SkParagraph

https://github.com/iPlug2/iPlug2/issues/508

### DIFF
--- a/Dependencies/IGraphics/NanoVG/README.md
+++ b/Dependencies/IGraphics/NanoVG/README.md
@@ -1,3 +1,5 @@
+*This project is not actively maintained.*
+
 NanoVG
 ==========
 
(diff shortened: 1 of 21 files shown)


AlexHarker:
I've been thinking about this recently (I've just made a plug where I do the EM ratio conversion the opposite way to set text sizes in the way that common image programs would), and I think this is definitely something that should be possible, but I really think we need a flag to go back to the old behaviour. Accessing the EM Ratio stuff that underlies this is non-trivial.

As an alternative there is another route where we provide a facility to calculate the size to be equivalent to the old way as a helper - this would perhaps be neater, but I think the decision process needs to take into account:

- how annoying it will be when everyone's UIs look wrong, and how to minimise time spent fixing it
- how much we want mess/stuff in the codebase that relates to the decision to follow the old NanoVG approach

Would be good to chat and devise an approach - or perhaps get opinions more widely - not sure if we want to open that can of worms or not.....

PR #1170, opened 2024-12-07
App type macro and function

aidv:
Makes it possible to detect if the app is running as au2, vst3 or standalone using macro RUNNING_AS or with static function IPlugAPIBase::isRunningAs()


olilarkin:
Thanks for contributing, however we already have [IPlugAPIBase::GetAPI()](https://iplug2.github.io/docs/class_i_plugin_base.html#ae54ecac578087ac61dfb2ee60e46dc94) for that.

```
if (GetAPI() == EAPI::kAPIAPP)
{
// do app specific thing
}
``

Also: For future reference PRs should be atomic, and include only changes related to one feature.

PR #1178, opened 2024-12-17
code to pass spacebar keypress to DAW from plugin window

### LINKED ISSUE #1177: Spacebar is not passed to DAW on Windows when using the webview
**Describe the bug**
On windows the webview captures all the keypress including spacebar this prevents the DAW from starting and stopping playback when the plugin window has the focus

**To Reproduce**
Open a webview based editor on any DAW on windows, press the spacebar

**Expected behaviour**
The Daw should start playback even if the plugin window has the focus

**IMPORTANT DETAILS**
* ALL formats 
* Windows
* Webview

hor-net:
Fixes #1177


olilarkin:
Thanks for the PR, but this approach is too limited. Let's instead do it this way: https://github.com/iPlug2/iPlug2/pull/1181

PR #1180, opened 2024-12-17
Webview/Windows: fix resizing and scaling

### LINKED ISSUE #1179: Incorrect window size on Windows when using webview
**Describe the bug**
When using the webview for the UI the initial window size is right on windows, subsequent windows resize exponentially increases the window size. Also the webview scale is incorrect.

**To Reproduce**
Open the UI of a webview based plugin, the content is too large for the window. Then try to resize the window, if you close and re open the window it is much larger than what it should be

**Expected behaviour**
the webview should just fill the window area, also the size of the window should be consistent between each close/open

**IMPORTANT DETAILS**
* All formats
* Windows
* Webview

hor-net:
fixes #1179

### DIFF
--- a/IPlug/Extras/WebView/IPlugWebViewEditorDelegate.cpp
+++ b/IPlug/Extras/WebView/IPlugWebViewEditorDelegate.cpp
@@ -49,21 +49,19 @@ WebViewEditorDelegate::~WebViewEditorDelegate()
 
 void* WebViewEditorDelegate::OpenWindow(void* pParent)
 {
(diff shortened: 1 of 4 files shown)


olilarkin:
Thanks for this PR. Due to some misunderstanding/the insane complexity of high DPI support on windows, i am currently reworking the highDPI support of the non-igraphics examples, and will need to do this a bit differently

hor-net:
I have it working right now in my branch if you want to take a look!

SAverio

Issue #1194, opened 2025-02-11
AUv2 parameter updates not sample accurate


svantana:
sampleOffsets from the host are not sent through to the plugin from IPlugAU::SetParamProc.

olilarkin:
👍

PR #1195, opened 2025-02-11
AUv2 sample accurate parameters

svantana:
solves issue #1194

### DIFF
--- a/IGraphics/Platforms/IGraphicsMac_view.mm
+++ b/IGraphics/Platforms/IGraphicsMac_view.mm
@@ -132,6 +132,10 @@ static int MacKeyEventToVK(NSEvent* pEvent, int& flag)
   return code;
 }
 
+#ifndef IGRAPHICS_MENU_RCVR
+#warning The iPlug2 Obj-C namespace is not customized. Did you forget to include IPlugOBJCPrefix.pch?
+#endif
+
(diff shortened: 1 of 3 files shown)


TBProAudio:
It seems that the fix does not automatically enables sample accurate automation. Just tested with Logic and Reaper. Is there anything else I should do in IPlug to get sample accurate automation working? Thank you.

svantana:
> It seems that the fix does not automatically enables sample accurate automation. Just tested with Logic and Reaper. Is there anything else I should do in IPlug to get sample accurate automation working? Thank you.

Yes, you need to use the offset information manually in your DSP, same as with midi events. Or are you saying that the offsets are wrong?

svantana:
> Thank you. Values are correct, but offset is always 0.
> 
> If I trace it down, only IPlugAU::DoSetParameter is called. But I think IPlugAU::DoScheduleParameters should be called by DAW and then IPlugAU::DoSetParameter with correct buffer offset?

Oh dear, just tested in reaper-mac now, and it's pretty bad. AUv2 always gets offset=0 and VST3 always gets offset=buffersize-1. I had only tested in ableton previously, where it worked perfectly for both formats.

TBProAudio:
Yes, Logic and Reaper do not work. What about Ableton?

In any case I cannot find any valid documentation on the internet regarding this...

PR #1196, opened 2025-02-12
GetMouseIsDown() function in IGraphics

svantana:
IGraphics can report the mouse position and latest mouseDown position, but not whether mouse is actually currently down. This is a pretty small addition, but useful nonetheless. I imagine in the future one could add a more complete touch/mouse state interface, but this is a good start and very compatible with the existing functions.

### DIFF
--- a/IGraphics/IGraphics.cpp
+++ b/IGraphics/IGraphics.cpp
@@ -986,6 +986,7 @@ void IGraphics::OnMouseDown(const std::vector<IMouseInfo>& points)
   {
     mMouseDownX = points[0].x;
     mMouseDownY = points[0].y;
+    mMouseIsDown = true;
   }
 
   for (auto& point : points)
@@ -1065,6 +1066,7 @@ void IGraphics::OnMouseDown(const std::vector<IMouseInfo>& points)
 void IGraphics::OnMouseUp(const std::vector<IMouseInfo>& points)
 {
 //  Trace("IGraphics::OnMouseUp", __LINE__, "x:%0.2f, y:%0.2f, mod:LRSCA: %i%i%i%i%i", x, y, mod.L, mod.R, mod.S, mod.C, mod.A);
+  mMouseIsDown = false;
   
   if (ControlIsCaptured())
   {
--- a/IGraphics/IGraphics.h
+++ b/IGraphics/IGraphics.h
@@ -1597,6 +1597,9 @@ public:
    * @param y Where the Y position will be stored */
   void GetMouseDownPoint(float& x, float&y) const { x = mMouseDownX; y = mMouseDownY; }
   
+  /** Get the current mouse down state. */
+  bool GetMouseIsDown() const { return mMouseIsDown; }
+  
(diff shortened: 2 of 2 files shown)


etheory:
This is possible to do yourself, there is also an example here: https://github.com/iPlug2/iPlug2/blob/6c36a4807ecb9ed361e4f42004a1fbedf1ee2b4e/IGraphics/Controls/IControls.cpp#L1108-L1124 where you can see the mMouseDown value is what you are after.

svantana:
> This is possible to do yourself, there is also an example here:

Sure, but the idea here is to see the global mouse state. Compare to IGraphics::GetMouseDownPoint() - that one is also doable inside each component, but that itself doesn't cover all use cases.

etheory:
But again, @svantana you could quite easily add that yourself and make a PR for it? It'd just be a new function probably on IGraphics. If you need a hand I could help.

IGraphics/IGraphics.cpp
```diff
@@ -986,6 +986,7 @@ void IGraphics::OnMouseDown(const std::vector<IMouseInfo>& points)
   {
     mMouseDownX = points[0].x;
     mMouseDownY = points[0].y;
+    mMouseIsDown = true;
```
etheory, line 989: Why not just put this at the top of this function? Why does it matter is there is a single point or multiple?

etheory:
Are you able to elaborate on the specific use case that brought on this PR? I'm genuinely curious.

IGraphics/IGraphics.cpp
etheory: Also mMouseDown should be made private with a const accessor function.

etheory, line 989: Why isn't this at the top of the function? Why does it matter if there are multiple points or one?

svantana: I was just following the existing structure for x and y. In the case of multitouch, "mouseIsDown" is not well defined. A better solution would be to query the mouse state from the OS, but that's much more complex.

svantana:
@etheory my use case is mac-style help overlays - the help overlay only shows if you mouse-hover on a widget without clicking for 1 second. And there is no mouseDown info to get from the IControl interface.

olilarkin:
@AlexHarker what are your thoughts on this?

AlexHarker:
I can see how this might be useful, and I see an argument that multi-touch is hard to define. However, this implementation does not treat things in a balanced manner.

For touch if one point goes down this will return true. If *any* points go up then this will return false. Arguably this is also the case for mMouseDownX and mMouseDownY. Multiple points go down together and it doesn't change - it only changes when one point goes down on its own (which could be a mouse or just one finger).

Sorry - that's not super helpful but it would be nice for this to be neat. Would be prepared to look at implementing this alongside reporting of what has been handled for multitouch (and single) events but we'd need to agree on what this should tell us. Should it tell us:

- only one point is down
- only one point is down that went down when no other points were down and has not yet been released
- one or more points is down
- one point went down on it's own and has not yet gone up?
- something else?

PR #1203, opened 2025-02-25
Add new InitFrequencyAutoRange initialization function

etheory:
With reference to: https://iplug2.discourse.group/t/automatic-hz-khz-change-during-frequency-parameter-changes/984/3 add a new function called: InitFrequencyAutoRange that provides a frequency parameter that automatically changes formatting between Hz and kHz.

Uses the function:
```c++
  auto freqFormatter = [](double val, WDL_String& str, IParam* parent)
  {
    const double onethousand = parent->FromNormalized(parent->ToNormalized(1000.));
    const bool gte1k = val >= onethousand || val >= 1000.;
    const char* units = gte1k? "kHz" : "Hz";
    if (gte1k) val *= 0.001;
    const int displayPrecision = parent->GetDisplayPrecision();
    str.SetFormatted(MAX_PARAM_DISPLAY_LEN, "%.*f %s", displayPrecision, val, units);
  };
```

In order to both provide formatting, and also respect either the precomputed precision, or a call to:
`SetDisplayPrecision`

All feedback welcome.

### DIFF
--- a/IPlug/IPlugParameter.cpp
+++ b/IPlug/IPlugParameter.cpp
@@ -175,6 +175,21 @@ void IParam::InitFrequency(const char *name, double defaultVal, double minVal, d
   InitDouble(name, defaultVal, minVal, maxVal, step, "Hz", flags, group, ShapeExp(), kUnitFrequency);
 }
 
+void IParam::InitFrequencyAutoRange(const char* name, double defaultVal, double minVal, double maxVal, double step, int flags, const char* group)
+{
+  InitDouble(name, defaultVal, minVal, maxVal, step, "", flags, group, ShapeExp(), kUnitCustom);
+  auto freqFormatter = [](double val, WDL_String& str, IParam* parent)
+  {
+    const double onethousand = parent->FromNormalized(parent->ToNormalized(1000.));
+    const bool gte1k = val >= onethousand || val >= 1000.;
+    const char* units = gte1k? "kHz" : "Hz";
+    if (gte1k) val *= 0.001;
+    const int displayPrecision = parent->GetDisplayPrecision();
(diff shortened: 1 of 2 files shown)


IPlug/IPlugParameter.cpp
```diff
@@ -175,6 +175,21 @@ void IParam::InitFrequency(const char *name, double defaultVal, double minVal, d
   InitDouble(name, defaultVal, minVal, maxVal, step, "Hz", flags, group, ShapeExp(), kUnitFrequency);
 }
 
+void IParam::InitFrequencyAutoRange(const char* name, double defaultVal, double minVal, double maxVal, double step, int flags, const char* group)
+{
+  InitDouble(name, defaultVal, minVal, maxVal, step, "", flags, group, ShapeExp(), kUnitCustom);
+  auto freqFormatter = [](double val, WDL_String& str, IParam* parent)
+  {
+    const double onethousand = parent->FromNormalized(parent->ToNormalized(1000.));
```
etheory, line 183: This has to be added due to the imprecision of normalization causing an entry of 1000. using a `PromptUserInput` call as with `IVKnobControl`.

TBProAudio: Thanks for this. I don't want to be picky, but shouldn't there be a solution for the text input '1.2k' as well?

etheory: That would be inconsistent with every existing control but its certainly possible. Do you want me to add that option?

etheory: @TBProAudio here is a partially working implementation of what you asked: https://github.com/iPlug2/iPlug2/pull/1204 - it needs a bit more work and feedback, but it's definitely heading in the right direction.

If you use it in conjunction with: `SetPromptShowsParamLabel(true);` it's already quite compelling.

TBProAudio:
Thanks. Yes, works as intended.

PR #1209, opened 2025-03-19
Revert "VST3: IPlugVST3View reports screen-scaled dimensions"

olilarkin:
Reverts iPlug2/iPlug2#1158 which caused a regression with VST3 editor sizing on windows.

WebView UI plugins will not get the correct dimensions without this. If you need that functionality, use the work on branch VST3/screen-scaled . Hopefully soon i can find a solution that works everywhere


olilarkin:
@AlexHarker if you have a moment could do with your screen scaling knowledge to help fix an issue i have with the non igraphics ui dimensions on windows. Can show you tomorrow if you have time

AlexHarker:
Travelling tomorrow but later in the day could work.

Issue #1212, opened 2025-03-23
Macos IPopmenu items are not utf8 encoded


svantana:
When adding an item with non-ascii characters (e.g. "motörhead") to an IPopmenu, it gets garbled in the macos modal popup.

I'm pretty sure this edit accidentally removed the utf8 encoding:

https://github.com/iPlug2/iPlug2/commit/e7bbeb7ec3779c9f7a46cdd187dae3294dbf321d

olilarkin:
thanks https://github.com/iPlug2/iPlug2/commit/3254453085e82aa08076d199182c82206355cff0

PR #1213, opened 2025-03-23
Macos NSString inited with UTF8 (fixes issue #1212)

svantana:
A simple fix for an issue that was most likely a mistake to begin with


olilarkin:
thanks, fixed in https://github.com/iPlug2/iPlug2/commit/3254453085e82aa08076d199182c82206355cff0

PR #1217, opened 2025-04-17
Mousewheel handling reporting

AlexHarker:
This is an enhancement to make IGraphics report whether a mousewheel event is handled or not (as with double clicks and mouseovers)


olilarkin (approved):
LGTM

PR #1219, opened 2025-04-17
Make sure newly added parameters are not zeroed with UnserializeParams()

AlexHarker:
Although the loop here already exits early if the chunk has no more data it will still write one parameter with a zero (overwriting the default or current value) if the number of parameters has increased and an old plugin state is loaded.

This change just checks if something (hopefully) valid was read from the chunk before setting the parameter


olilarkin (approved):
Sounds reasonable. Nit: insert space after >=

Issue #740, opened 2021-05-11
VST3: Responsive UI example is the wrong size on windows


olilarkin:
In Reaper, on a high DPI screen, it seems 2x as big as it should be

olilarkin:
fix on https://github.com/iPlug2/iPlug2/tree/VST3/screen-scaled

see also

https://github.com/iPlug2/iPlug2OOS/issues/10

olilarkin:
@AlexHarker would like to chat about this soon, since IIRC you weren't convinced there was a problem with the VST3 view size reporting

olilarkin:
This should be fixed by https://github.com/iPlug2/iPlug2/pull/1158

If problems emerge will revert it

PR #1220, opened 2025-04-17
Fix scaling of responsive UI IGraphics plugins on windows

olilarkin:
This should be the correct fix for https://github.com/iPlug2/iPlug2/issues/740

### DIFF
--- a/Examples/IPlugResponsiveUI/IPlugResponsiveUI.cpp
+++ b/Examples/IPlugResponsiveUI/IPlugResponsiveUI.cpp
@@ -59,11 +59,6 @@ IPlugResponsiveUI::IPlugResponsiveUI(const InstanceInfo& info)
 }
 
 #if IPLUG_EDITOR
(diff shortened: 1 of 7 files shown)


AlexHarker:
Looks like we can still get an onSize() call to the VST3 view in REAPER even if we've previously return false for canResize.

I'm wondering is onSize needs to examine where host resizing is on or not.

AlexHarker:
BTW - this affects all VST3 plugin builds, as they will get resized (at least in REAPER)

olilarkin:
I see. I am adding a check to onSize(). I am surprised that reaper still sends onSize() since we should have returned false to canResize()

olilarkin:
https://github.com/iPlug2/iPlug2/pull/1221

AlexHarker:
Me too, but that's what happens - new PR looks good on a skim

AlexHarker:
REAPER only then calls to constrain the size if the plugin has reported that it can resize.

PR #1223, opened 2025-04-19
Fix resource bundling on Xcode 16.3

### LINKED ISSUE #1222: XCode 16.3 breaks example projects
Some changes with XCode 16.3 mean that plugin resources are not correctly bundled. It seems that the TARGET_BUILD_DIR environment variable no longer has a path eliminator on the end causing the python script to place resources outside the bundle

olilarkin:
Fixes #1222


olilarkin:
Seems to still work with < Xcode 16.3

normalaccount2137 (commented):
It indeed fixes the issue where /Resources were not correctly bundled within /Contents upon compiling with Xcode 16.3

PR #1229, opened 2025-04-20
IGraphicsIOS: Add HoverGestureRecognizer for mouse-overs

olilarkin:
Adds a [uihovergesturerecognizer](https://developer.apple.com/documentation/uikit/uihovergesturerecognizer)

When an iPad / mac Catalyst app / visionOS app receives pointer info we currently don't issue any mouse overs. This fixes that. I believe that the gesture recogniser is freed via ARC, and that on e.g. iPhone where there are not pointer events it shouldn't have any negative impact on performance

### DIFF
--- a/IGraphics/Platforms/IGraphicsIOS_view.mm
+++ b/IGraphics/Platforms/IGraphicsIOS_view.mm
@@ -230,6 +230,10 @@ extern StaticStorage<CoreTextFontDescriptor> sFontDescriptorCache;
   [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationWillEnterForegroundNotification:) name:UIApplicationWillEnterForegroundNotification object:nil];
   mColorPickerHandlerFunc = nullptr;
   
+  UIHoverGestureRecognizer* hoverGestureRecognizer =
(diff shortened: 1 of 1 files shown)


IGraphics/Platforms/IGraphicsIOS_view.mm
```diff
@@ -832,6 +836,22 @@ - (void) onRotateGesture: (UIRotationGestureRecognizer*) recognizer
...
 
+- (void) onHoverGesture: (UIHoverGestureRecognizer*) recognizer
+{
+  CGPoint pos = [recognizer locationInView:self];
+
+  IMouseInfo info;
+  
+  auto ds = mGraphics->GetDrawScale();
+  
+  info.ms.L = true;
```
AlexHarker, line 847: Is there a reason to set this to true? Normally both L and R would be up when getting mouseovers?

At some point we should also add keyboard input (there's something on a branch somewhere) and set other modifiers correctly.

AlexHarker (changes requested):
Can you check the mouse state code (and correct if needed).

Otherwise (assuming this is tested) it looks worthwhile and correct!

PR #1232, opened 2025-04-20
Add ScopedGLContext

olilarkin:
This is a highly risky PR that I would like to get merged before a subsequent PR that would merge GLES support via ANGLE.

Rationale, for ANGLE: Apple have deprecated OpenGL and we never supported it on iOS in iPlug2. I notice some strange artefacts with IGRAPHICS_GL3 / IGRAPHICS_NANOVG on my Apple Silicon mac when I run IGraphicsTest. I am not sure if this is due to the way we create the GL Context (there are different ways) or if something is just broken with GL on macOS.

Anyhow, I would like to add GLES support via ANGLE, which is the library that allows WebGL to work consistently across many browsers. That means someone can use IGraphics with GLES on all platforms.

This PR is the first part of that work to clean up the Activate/Deactivate/SwapBuffers calls. It's a bit fiddly so I could do with some more 👀

### DIFF
--- a/IGraphics/Drawing/IGraphicsNanoVG.cpp
+++ b/IGraphics/Drawing/IGraphicsNanoVG.cpp
@@ -331,18 +331,16 @@ APIBitmap* IGraphicsNanoVG::LoadAPIBitmap(const char* fileNameOrResID, int scale
 
     if (pResData)
     {
-      ActivateGLContext(); // no-op on non WIN/GL
+      ScopedGLContext scopedGLCtx {this};
       idx = nvgCreateImageMem(mVG, nvgImageFlags, (unsigned char*) pResData, size);
-      DeactivateGLContext(); // no-op on non WIN/GL
     }
   }
   else
 #endif
   if (location == EResourceLocation::kAbsolutePath)
   {
-    ActivateGLContext(); // no-op on non WIN/GL
+    ScopedGLContext scopedGLCtx {this};
     idx = nvgCreateImage(mVG, fileNameOrResID, nvgImageFlags);
-    DeactivateGLContext(); // no-op on non WIN/GL
   }
 
   return new Bitmap(mVG, fileNameOrResID, scale, idx, location == EResourceLocation::kPreloadedTexture);
@@ -358,10 +356,11 @@ APIBitmap* IGraphicsNanoVG::LoadAPIBitmap(const char* name, const void* pData, i
     int idx = 0;
     int nvgImageFlags = 0;
 
-    ActivateGLContext();
-    idx = nvgCreateImageMem(mVG, nvgImageFlags, (unsigned char*)pData, dataSize);
-    DeactivateGLContext();
-
+    {
+      ScopedGLContext scopedGLCtx {this};
+      idx = nvgCreateImageMem(mVG, nvgImageFlags, (unsigned char*)pData, dataSize);
+    }
+    
     pBitmap = new Bitmap(mVG, name, scale, idx, false);
 
     storage.Add(pBitmap, name, scale);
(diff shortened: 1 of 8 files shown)


olilarkin:
@svantana would you be able to take a look at this? You are using IGRAPHICS_GL3 right?

IGraphics/IGraphics.h
```diff
@@ -970,6 +970,17 @@ class IGraphics
...
+  // An RAII helper to manage the IGraphics GL context
+  class ScopedGLContext
+  {
+  public:
+    ScopedGLContext(IGraphics* pThis)
+    : mThis(pThis) { mThis->ActivateGLContext(); }
+    ~ScopedGLContext() { mThis->DeactivateGLContext(); }
+  private:
+    IGraphics* mThis;
+  };
```
AlexHarker, line 982: This looks like a good idea, but I'd avoid using "this" as a member/parameter naming, as it could also be used from a Mac view correctly.

As this object also makes no sense if there is no IGraphics I'd suggest it should be stored by reference to the object, rather than by pointer.

IGraphics/Drawing/IGraphicsNanoVG.cpp
AlexHarker, line 1: No changes in this file look risky or controversial to me - cleaner and better I'd say!

IGraphics/Platforms/IGraphicsWin.cpp
```diff
@@ -937,15 +932,6 @@ void IGraphicsWin::PlatformResize(bool parentHasResized)
   }
 }
 
-#ifdef IGRAPHICS_GL
```
AlexHarker, line 940: Is there no longer a need to activate and deactivate around draw resizing or is it happening some other way?

IGraphics/Platforms/IGraphicsWin.cpp
AlexHarker, line 1: Other than checking on DrawResize() this looks good to me and again cleaner and less susceptible to issues.

Minor point -  I'm not a massive fan of scoping using blocks with no control flow (ifs/switches etc.) but I can see why that has been done here. Might be worth thinking about.

IGraphics/Platforms/IGraphicsWin.h
AlexHarker, line 1: Nothing here that isn't covered elsewhere - presume the moving of method visibility is intentional

AlexHarker:
Looked through and added comments - not done the Mac stuff yet, but mostly good - some caution of checking needed around DrawResize() - happy to discuss sometimes if desirable.

IGraphics/Platforms/IGraphicsWin.cpp
olilarkin: It happens in NanoVG DrawResize() but i see that we also need it in Skia DrawResize

IGraphics/Platforms/IGraphicsWin.h
olilarkin: yeah

IGraphics/Platforms/IGraphicsWin.cpp
olilarkin: I added it to skia

IGraphics/IGraphics.h
olilarkin: I tweaked

olilarkin (commented):
Didn't mean to start a review on my own PR.

svantana:
> @svantana would you be able to take a look at this? You are using IGRAPHICS_GL3 right?

Ok, finally got around to giving this a spin on mac and (VM) windows. I tried the things that have been problems in the past, which are:

* multiple instances
* moving to and from external monitor
* opening and closing window very fast

All seem to work without issue. Yay!

IGraphics/Platforms/IGraphicsWin.cpp
olilarkin: I believe this is addressed by adding the scoped context to IGraphcsiSkia:: DrawResize()

AlexHarker:
This was slightly broken for non-GL builds due to the positioning of a preprocessor directive, leading to linking errors.

I pushed the fix directly to the main branch, as although I don't like doing that we had a broken state on that branch, and I had tested locally.

olilarkin:
Thanks!

PR #1238, opened 2025-04-20
Replace Win 32 builds with ARM64EC

olilarkin:
This PR swaps out the (obsolete) 32 bit windows builds for ARM64EC targets.

common.props variables that were 

XX_32 ->  XX_ARM64EC 
XX_64 -> XX_x86_64


sstillwell:
RE: AAX - Pro Tools won't run on ARM64EC (primarily due to the fact that PACE tools won't run on ARM64EC, so at least in my repos (I'm already building ARM64EC versions), I just don't build AAX and stick to VST2/VST3.

PR #1244, opened 2025-04-21
WIP: Add support for visionOS

olilarkin:
Thanks to a lot of hard work on iOS/AUv3 support over the years, iPlug2 is ready to enter the age of spatial computing 🚀


hor-net:
That’s great Oli, i was talking with an Apple rep some time ago and they are searching for apps for VisionOS (now we just need a killer idea :D)

Saverio

PR #1246, opened 2025-04-25
Fix VST2 casting so that the compiler gives no warnings

AlexHarker:
There have been some warnings in the VST2 main function due that can be fixed by casting in an appropriate C++ style to let the compiler know that the casts are fully intentional. This PR does that.


olilarkin:
Nice

olilarkin (approved):
LGTM

PR #1247, opened 2025-04-26
Skia image filter dropshadow

AlexHarker:
This is an optimisation/enhancement.

This provides a way that Skia GPU backends can use an image filter to do shadows that have a single color, leading to currently very similar results visually as the convolution drop shadow in IGraphics with much improved performance. This also allows drop shadows to run on IOS where grabbing the pixels fails (or crashes I can't quite remember) for security reasons.

Notes:
- **I would like to tweak the approximation of E** to see if we can get even more similar outputs for the same input **before we merge**.
- The IGraphics code is used for gradient pattern shadows to ensure there is no change in functionality (this is not supported by skia image filters and it's not obvious that we could do this in skia with a hand-rolled filter
- On the CPU there is not evidence that using an image filter is faster, and some cases where it is slower, as the code in IGraphics tries to determine a smaller area for the shadow which will sometimes be the case.
- We can squash this to a single merge without needing to keep any of the history.

### DIFF
--- a/IGraphics/Drawing/IGraphicsSkia.cpp
+++ b/IGraphics/Drawing/IGraphicsSkia.cpp
@@ -17,6 +17,7 @@
 #include "include/core/SkPathEffect.h"
 #include "include/effects/SkDashPathEffect.h"
 #include "include/effects/SkGradientShader.h"
+#include "include/effects/SkImageFilters.h"
 
 #if !defined IGRAPHICS_NO_SKIA_SKPARAGRAPH
 #include "modules/skparagraph/include/FontCollection.h"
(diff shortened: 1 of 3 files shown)


AlexHarker:
I've looked again at the scaling and managed to match the current IGraphics blur amounts with a constant factor of 3.f

I'm not sure why this is the value - it looks like we've inherited whatever scaling I used in much earlier code, but I don't want to break the look of anyone's plugins. 

To show the comparison (which I think looks indistinguishable see the below. The blur here is hitting the edges of the layer, but that's not a bug - just how it works, and it helps to check that things look the same in both cases.

Screenshot 2025-04-27 at 10 14 09
Screenshot 2025-04-27 at 10 14 12

olilarkin:
Seems like a good improvement. Does the image filter also work for SKIA CPU? 

It looks a tad darker on my machine:
image

If this PR is not yet for merging, please make it a draft PR.

AlexHarker:
I thought is was ready to merge, but it doesn't look to be. I'm confused though as to why the blurring is so different on your version to mine - it's much tighter on Skia METAL which will also lead to the darker effect. Let me check that test.

AlexHarker:
OK - @olilarkin could you give this another spin - I wasn't accounting for opacity, which luckily is part of the test (but not my custom test above).

AlexHarker:
>  Does the image filter also work for SKIA CPU?

Yes, but as per the notes above it won't be called if you are on CPU because it isn't faster than the IGraphics code and can in specific circumstances be (a lot) slower (if your layer is mostly empty IGraphics does some optimisation on the convolution to avoid unneeded calculations).

olilarkin:
Looks pretty much the same now

image

AlexHarker:
Cool - to me those two windows look different colours - not sure if that's a capture thing, or if that's the same on the screen - might be a color profile related issue. I tend to launch both (using add) and then flip between and here on my screen it's very hard to tell the difference.

olilarkin:
I squashed the commits and force-pushed. made a misttake... fixing from my backup

olilarkin:
LGTM, should be good to go now

olilarkin (approved):
LGTM

PR #1249, opened 2025-05-03
Separate some of the windows font code from IGraphicsWin (.h/,cpp) 

AlexHarker:
This code means that font code in IGraphicsWin that is currently private can be reused for different platform implementations on windows, rather than creating a need to duplicate (the Mac fonts code is already separated). There should be no change to functionality.

It does create a new header, so whilst projects (Examples etc.) do not require changes to compile, it would be good to update them in order to include this file in projects in the future.

@olilarkin - do you have a nice way to update examples etc. in cases like this, rather than working fully manually? I'm also aware this should potentially go back to OOS.


olilarkin:
I don't unfortunately, but we can update the projects in the future like you said.

olilarkin (approved):
LGTM

PR #1254, opened 2025-05-04
Adapt IGraphicsTest for responsive UI

### LINKED ISSUE #1253: IGraphicsTest is the wrong dimensions on iOS
Since https://github.com/iPlug2/iPlug2/commit/0c4e74b33f915764bae8c88a3c666e5002a3de51

IGraphicsTest is not working well on iOS. We either need to disable PLUG_HOST_RESIZE for iOS or adapt the UI layout to be adaptive. I would vote for the latter, also to improve the test with an IVTabbedPagesControl

olilarkin:
fixes #1253 

https://github.com/user-attachments/assets/90fe535d-f0ba-4e6b-b585-db5db4cf7a44


olilarkin:
@AlexHarker what do you think of this change in principle?

PR #1255, opened 2025-05-04
Remove IGraphicsCanvas

olilarkin:
In order to reduce the complexity of IGraphics implementations, let's remove IGraphicsCanvas, a little used backend for iPlug2 WAMs. NanoVG can be used via WebGL, and maybe Skia in the future


AlexHarker (approved):
LGTM

PR #1257, opened 2025-05-10
Corner resizer redraw after resting is finished

AlexHarker:
In resize mode when drag resizing finishes the oorner resizer is not currently made dirty. This is not an issue for rescaling as all controls are made dirty in that scenario.

This PR fixes that as the corner resizer may need to draw differently when resizing is not taking place.


olilarkin (approved):
LGTM

PR #1342, opened 2026-04-05
make c++ standard and AAX SDK folder paths configurable by cmake

pressplay-music:
Make the two cmake variables `IPLUG2_CXX_STANDARD` and `AAX_SDK_DIR` configurable.

These changes should be non-breaking for existing cmake configurations. They allow you to set a newer c++ standard (was c++20 in my case) and a custom out-of-source AAX SDK path (which is good because you will want to define your own AAX cmake scripts and not commit them to iplug2).

### DIFF
--- a/Scripts/cmake/AAX.cmake
+++ b/Scripts/cmake/AAX.cmake
@@ -12,7 +12,9 @@ include(${CMAKE_CURRENT_LIST_DIR}/IPlug.cmake)
 
 if(NOT TARGET iPlug2::AAX)
   # Define SDK path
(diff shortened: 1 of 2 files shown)


pressplay-music:
I don't know the reason why it is failing. Perhaps because I disabled the "allow edits" setting at first. But even after enabling it and force-pushing the error persists.

olilarkin:
Superseded by #1372. The `IPLUG2_CXX_STANDARD` half of this PR landed via #1371; #1372 picks up the remaining `AAX_SDK_DIR` change. Thanks @pressplay-music!

Issue #1366, opened 2026-05-27
cmake CXX standard is forced to 17


svantana:
A custom IPLUG2_CXX_STANDARD is overridden in iPlug2.cmake (line 36)

PR #1367, opened 2026-05-27
cmake: only set CXX standard if not defined

svantana:
Fix for issue #1366


olilarkin:
Thanks, will replace with #1371

PR #1387, opened 2026-07-30
IGraphicsMac: fix CAMetalLayer leak on window close

olilarkin:
makeBackingLayer returned a +1 layer from a non-ARC file; AppKit retains the returned layer, so it was stranded (with its ~framebuffer-sized drawable pool) on every window close. Also balance the Create-rule +1 from MTLCreateSystemDefaultDevice(). Metal analogue of the GL fix in PR #1061 (e4a0d2a79).


olilarkin:
@AlexHarker what do you think about merging this

PR #1388, opened 2026-07-31
ReaperExt: add WEB_RESOURCES_DIR to iplug_configure_reaperext()

olilarkin:
A REAPER extension is a bare dylib/DLL with no bundle to carry resources, so an extension with a WebView UI has to deploy `resources/web` next to itself for release builds to find `index.html` at runtime.

`IPlugReaperExtensionWebUI` did that with fifteen lines of `add_custom_command` in its own `CMakeLists.txt`. That is boilerplate every WebView extension needs, and it is easy to lose: a project derived from the example by copying the sources builds and runs fine in debug, then shows "file not found" in release, because the missing piece lives in the build file rather than the code. That is exactly how it went missing in a downstream template, which is what prompted this.

This hoists it into `iplug_configure_reaperext()` behind an optional `WEB_RESOURCES_DIR` argument, and converts the example to use it:

```cmake
iplug_configure_reaperext(${PROJECT_NAME} ${PROJECT_NAME}
  WEB_RESOURCES_DIR ${PROJECT_DIR}/resources/web
)
```

The destination is `<UserPlugins>/<project_name>/`, resolved through `iplug_get_default_deploy_path(REAPEREXT)`, so it honours `IPLUG_REAPEREXT_DEPLOY_PATH` overrides and lands correctly on macOS as well as Windows. It has to agree with `SHARED_RESOURCES_SUBPATH` in `config.h`, which also defaults to the project name.

A relative `WEB_RESOURCES_DIR` is resolved against the caller's source dir, since `POST_BUILD` commands run in the binary dir, and a directory that does not exist is reported at configure time rather than at the end of the first build.

## Testing

Windows, VS2022. `Examples/IPlugReaperExtensionWebUI` built Release with `-DIPLUG_REAPEREXT_DEPLOY_PATH=<scratch>` so the deploy could be inspected without touching an installed REAPER. Deployed layout is identical to what the hand-rolled block produced:

```
<deploy>/reaper_IPlugReaperExtensionWebUI.dll
<deploy>/IPlugReaperExtensionWebUI/index.html
<deploy>/IPlugReaperExtensionWebUI/script.js
```

Path handling checked both ways against a downstream project passing a relative directory:

- `WEB_RESOURCES_DIR resources/web` resolves to the source tree and configures clean
- `WEB_RESOURCES_DIR resources/nope` fails configure with `does not exist (looked in ...)`

Not yet exercised on macOS — the path comes from `iplug_get_default_deploy_path(REAPEREXT)`, which already resolves to `~/Library/Application Support/REAPER/UserPlugins` there, but that arm is untested.

### DIFF
--- a/Examples/IPlugReaperExtensionWebUI/CMakeLists.txt
+++ b/Examples/IPlugReaperExtensionWebUI/CMakeLists.txt
@@ -55,20 +55,10 @@ iplug_add_target(${PROJECT_NAME} PUBLIC
 )
 
 # Configure as REAPER extension (sets up output, deployment, etc.)
-iplug_configure_reaperext(${PROJECT_NAME} ${PROJECT_NAME})
-
-# Deploy the WebView UI next to the extension so Release builds can load index.html at runtime.
-# (In Debug the extension loads directly from the source tree via __FILE__.) The destination
-# subfolder must match SHARED_RESOURCES_SUBPATH in config.h.
-if(IPLUG_DEPLOY_PLUGINS)
-  iplug_get_default_deploy_path(REAPEREXT)
-  if(IPLUG_DEPLOY_PATH_REAPEREXT)
-    add_custom_command(TARGET ${PROJECT_NAME} POST_BUILD
-      COMMAND ${CMAKE_COMMAND} -E copy_directory
-        "${PROJECT_DIR}/resources/web"
-        "${IPLUG_DEPLOY_PATH_REAPEREXT}/IPlugReaperExtensionWebUI"
-      COMMENT "Deploying WebView UI to ${IPLUG_DEPLOY_PATH_REAPEREXT}/IPlugReaperExtensionWebUI"
-      VERBATIM
-    )
-  endif()
-endif()
+# WEB_RESOURCES_DIR deploys the WebView UI next to the extension so Release builds can load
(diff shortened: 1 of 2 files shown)


olilarkin:
Good catch — fixed in 245621c53.

You're right that `${PROJECT_DIR}` masked it: both in-tree callers pass absolute paths, so nothing exercised the relative case. Resolved with `get_filename_component(... ABSOLUTE BASE_DIR ${CMAKE_CURRENT_SOURCE_DIR})` rather than `WORKING_DIRECTORY`, so the resolution happens once at configure time and the value baked into the build rule is absolute either way.

Also added an `IS_DIRECTORY` guard. Without it a bad path fails at the end of the first build with a bare `copy_directory` error naming a path the caller never wrote; now it fails at configure time and reports both what was passed and where it was looked for.

Verified both arms on Windows with a relative `WEB_RESOURCES_DIR resources/web`:

- resolves to the source tree and configures clean
- `resources/nope` → `WEB_RESOURCES_DIR 'resources/nope' does not exist (looked in ...)`

The example keeps `${PROJECT_DIR}/resources/web`, matching the convention used for `INCLUDE` in every other example, but the relative form now works too.

olilarkin (commented):
LGTM

